PR #2, opened 2017-11-22
preparations for testing HDC backend

coobas:
These should be harmless changes (test suite works). There is a new dependence on the future package (which is very useful for Py 2/3 development).
Sorry for being a bit pe-p8-dantic - I simply follow the editor's linter complaints. Plus I believe pep8 is indeed useful :)


orso82:
@coobas it is a good point to apply pep8 now that we are at the early stages of development.
For consistency I have applied pep8 to all `.py` files (for reference Alt-Command-L on a Mac with Pycharm)

smithsp:
If we care enough about PEP8, we should put in a git commit hook.

orso82:
@smithsp I do not think we are pep8-OCD at that point yet :)))

PR #4, opened 2017-12-12
Added MANIFEST.in so license is included with pypi distribution

smithsp:
Please also upload a new version to pypi.

This is to support https://github.com/conda-forge/staged-recipes/pull/4601


orso82:
Done. Thank you @smithsp . Let me know if there is anything else that is needed.

PR #8, opened 2018-03-21
Feature/hdc

coobas:


omas/omas_setup.py
```diff
@@ -16,17 +16,15 @@
 import uncertainties
 import uncertainties.unumpy as unumpy
 from uncertainties.unumpy import nominal_values, std_devs, uarray
-import StringIO
+from io import StringIO
```
smithsp: If this merge is not done soon, then this change should be propagated to master, as it is causing a problem on the conda-forge build: https://circleci.com/gh/conda-forge/omas-feedstock/16?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

orso82: I will work on it.

coobas:
Github is slow ... We might be a bit slower with HDC@conda but we'll get there ;)

PR #10, opened 2018-03-29
Refactoring `omas --> OMAS`

orso82:
@gafusion/omas_developers this updates the OMAS API as follows:

```python
import omas
ods=omas.omas()
```

will become

```python
import omas
ods=omas.ODS()
```

and also changes `tokamak` to `machine`, so to be more consistent with IMAS nomenclature.


orso82:
Yes. And found many things that I missed on the first pass.

smithsp:
@orso82 Did you merge this, and GitHub is slow to record it?

orso82:
Actually I did not realize I was not pushing directly on master. I will be pushing few more fixes, merge this on master and publish v0.9.1

Thank you for pointing this out

PR #16, opened 2018-05-07
Added cocos_signals for core_profiles and core_sources

bclyons12:
Some details on decisions I made:

Phi has been updated to use BT direction, according to Eq. 45 of the COCOS reference paper:
screen shot 2018-05-07 at 10 32 31 am

Electric field and V_loop follow sign of IP (ensuring E=eta*J has positive resistivity)

Torque/momentum/velocities follow sign of BT (allowing for flip in toroidal direction)


orso82:
@bclyons12 awesome 👍 ! Thank you!

One more thing is that we should remove the `None`s from the list of COCOS signals. I just left it there because I wanted to check what the conversion factors were for some of those quantities, but we should remove them.

@sautero may be interested in checking out our work

bclyons12:
@orso82 a36ae7c fixes the None issue from your last comment

PR #17, opened 2018-05-08
Fix some cocos_signals

bclyons12:
-Some "b_field" entries are |B|, so no transform
-Radial components have no trasnform
-Poloidal components get new transform


orso82:
👍 @bclyons12

PR #23, opened 2018-06-06
efitviewer-style diagnostic and hardware overlays as method of ODS

eldond:
- ODS should be able to hold everything needed to make a tokamak cross section plot with overlays showing positions of various hardware systems
- OMFIT will provide utilities for loading minimal hardware info to enable these overlays; see gafusion/OMFIT-source#2791
- Overlays should also work fine on a fully developed and populated ODS
- Add diagnostic/hardware overlay plot to omas_plot
  - The master `overlay` function calls individual overlays depending on which are switched on
  - So far, the individual hardware systems with overlay methods are: Thomson, bolometers, and gas puffing
  - More hardware systems to be added later
- Add some utility functions in omas_utils

Changes needed/suggested (please edit as needed):
- [x] Individual hardware system overlays detect whether or not they have adequate data and abort themselves gracefully if they're missing something.
- [x] Generalize logic for choosing which overlays to call instead of giving each one of them its own `if`. See review by @smithsp .
- [x] More systems (I don't think all are needed to start with, but we should get all the essentials):
  - [x] CER overlay (with keywords to select which beams and view directions (tang vs vert))
  - [x] Magnetic probe overlay
  - [x] PF coil overlay
  - [x] Interferometer overlay
- [ ] Stress test custom keywords like `labelevery` and `mask`
- [ ] Final review and polish

omas_efitviewer_cx

### DIFF
--- a/omas/omas_plot.py
+++ b/omas/omas_plot.py
@@ -460,11 +460,16 @@ def overlay(ods, ax=None, **kw):
     """
     if ax is None:
         ax = pyplot.gca()
-    supported_systems = ['gas_injection', 'thomson_scattering', 'bolometer']
-    for hw_sys in supported_systems:
-        if kw.get(hw_sys, False):
+    overlay_on_by_default = ['thomson_scattering']  # List of strings describing default hardware to be shown
+    for hw_sys in list_structures(ods.imas_version):
+        if kw.get(hw_sys, hw_sys in overlay_on_by_default):
             overlay_kw = kw.get(hw_sys, {}) if isinstance(kw.get(hw_sys, {}), dict) else {}
-            eval('{}_overlay'.format(hw_sys))(ods, ax, **overlay_kw)
+            try:
+                overlay_function = eval('{}_overlay'.format(hw_sys))
+            except NameError:
+                pass
+            else:
+                overlay_function(ods, ax, **overlay_kw)
     return
 
 
--- a/omas/omas_plot.py
+++ b/omas/omas_plot.py
@@ -457,7 +457,7 @@ def overlay(ods, ax=None, **kw):
 
     :param ax: Axes instance or None
 
-    :param kw: select plots by setting their names to True; e.g.: if you want the gas_injection plot,
+    :param **kw: select plots by setting their names to True; e.g.: if you want the gas_injection plot,
         set gas_injection=True as a keyword.
         Instead of True to simply turn on an overlay, you can pass a dict of keywords to pass to a particular overlay
         method, as in thomson={'labelevery': 5}. After an overlay pops off its keywords, remaining keywords are passed
@@ -527,7 +527,7 @@ def gas_injection_overlay(ods, ax=None, angle_not_in_pipe_name=False, **kw):
         Set this to include (Angle) at the end of injector labels. Useful if injector/pipe names don't already include
         angles in them.
 
-    :param kw: Additional keywords for gas plot:
+    :param **kw: Additional keywords for gas plot:
         colors: List of colors for the various gas ports. The list will be repeated to make sure it is long enough.
             Do not specify a single RGB tuple by itself. However, a single tuple inside list is okay [(0.9, 0, 0, 0.9)]
         *Remaining keywords are passed to plot call for drawing markers at the gas locations.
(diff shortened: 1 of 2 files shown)


omas/omas_plot.py
```diff
@@ -446,3 +446,136 @@ def core_profiles_pressures(ods, time_index=0, ax=None, **kw):
...
+    if ax is None:
+        ax = pyplot.gca()
+    if kw.pop('gas', False):
+        gas_overlay(ods, ax)
+    if kw.get('thomson', False):
+        tskw = kw.get('thomson', {}) if isinstance(kw.get('thomson', {}), dict) else {}
+        thomson_overlay(ods, ax, **tskw)
+    if kw.get('bolometer', False):
+        bolkw = kw.get('bolometer', {}) if isinstance(kw.get('bolometer', {}), dict) else {}
+        bolometer_overlay(ods, ax, **bolkw)
```
smithsp: I suspect this function could be generalized to process `**kw`, such that the keys of **kw are the names of the IDS's, like
image
(with `name.replace(' ','_')`) and you can use something like
```python
for k,v in kw.items():
  k = k.replace(' ','_')
  try:
    eval(k+'_overlay)
  except NameError:
    print('%s does not exist'%(k+'_overlay'))
    print('Must be one of:')
    overlays = [d for d in dir() if d.endswith('_overlay')]
    print(overlays)
    # raise NotImplementedError ?
  else:
    processed_kw = v if isinstance(v,dict) else {}
    eval(k+'_overlay')(**processed_kw)
```

eldond: You are right.

orso82:
@eldond can you please add a newline in the doc-string between each `:parameter X: directive? Also, instead of `:param kw:` can you write it as `:param \**kw`

What you have now gets rendered like this:
image

With these changes, it will become:
image

eldond:
@orso82 Okay, I'm on it. Should I also put newlines between items within **kw?

eldond:
The ability to use `ods['thomson_scattering']['channel'][0]['position']['r']` interchangeably with `ods['thomson_scattering.channel.0position.r']` makes omas SO MUCH EASIER TO USE. There are statements that would've been so messy to implement that I wouldn't have bothered (or maybe would've made a huge mess) that are just trivial with this `.` or `']['` system. OMAS is so smooth to deal with so far.

eldond:
@orso82 What does this mean? Why are the units `m`? How do I enter this properly? My data setup function and plot function are self consistent and working, but I am wary that a change in convention might be needed. That's fine; it's easier to flip the sign or something in two places in a working system and check that it's still working.

Path | Dimensions | Type | Units | Description
-- | -- | -- | -- | --
pf_active.coil[:].element[:].geometry.oblique.alpha |   | FLT_0D | m | Inclination of first angle TBD



https://gafusion.github.io/omas/schema/schema_pf%20active.html

smithsp:
@eldond We should probably get you access to ITER so you can post these types of questions there.

eldond:
@orso82 I think this first set of overlays is ready for review as I do final checks. Right now, the test script is in OMFIT. How do you want me to go about adding tests to the omas repo itself? I see `ods_sample.py` is in there; can I use that to drive tests? Should I save another ods sample with all the coordinates and stuff in it? Should the test try to import OMFIT? That would fail on travis, though.

omas/omas_plot.py
```diff
@@ -446,3 +514,473 @@ def core_profiles_pressures(ods, time_index=0, ax=None, **kw):
...
+def magnetics_overlay(ods, ax=None, **kw):
+    """
+    Plots overlays of magnetics: B_pol probes and flux loops
+
+    :param ods: OMAS ODS instance
+
+    :param ax: Axes instance
+
+    :param \**kw: Additional keywords
+        show_bpol_probe: bool
```
orso82: @eldond I feel that some of these arguments should be explicit and not be in `**kw`. For the hardware specific overlay function (not the general one) I would leave in `**kw` only the plotting options that are passed to matplotlib.

eldond: Okay, I made most of them explicit. There were a few places where it felt more natural to keep them inside **kw, but I can change this if you like. Still in **kw:
 - general keywords that all the hardware overlays accept (`labelevery` and `mask`)
 - CER overlay colors and markers. I loop through strings and pop them. It seems nicer that way. I guess I could `eval` or loop through a `zip` if I made these keywords explicit.
 - scalex and scaley in `pf_active_overlay`: these are popped off and passed to `autoscale_view`. The user probably won't bother with them, so I want to try to tuck them away. The matplotlib function I grabbed this idea from did it this way.

eldond:
@orso82 I didn't put in the tests yet! It only is tested in OMFIT; there is no test within the OMAS repo.

orso82:
@eldond I have not yet had the time to start down the path of creating unit-tests within OMAS. My plan was to make a `tests` folder and set them up there. There are several scripts under `sample` and the `utilities/omas_suite_test.py` that will have to move there.

If you want you can start down that path, and I'd be happy to follow your lead.

eldond:
Okay. Do you want to set up something like codecov on the repo to measure test coverage? Do you want me to mess with your travis files, or just create tests that you can link as you like? I guess I could do edits in a PR and let you reject/cherry pick parts of it, assuming I get that far.

PR #25, opened 2018-06-13
Expand tests for omas_plot

eldond:
It turns out that writing better tests helps find more bugs.

Measuring test coverage helps find where tests need improvement.

This is a great system.


eldond:
@orso82 Ready for review/merge

PR #26, opened 2018-06-13
PF coil and gas arrow plots use cocos_transform

eldond:
- Cope with the possibility that the ODS is not in COCOS 11.
- Also reversed the convention for oblique angles in PF active (both here and in OMFIT). OMFIT loads them in COCOS 11. The calculations transform them to COCOS 11 using cocos_transform, then go about their business.
- Internally double-reversed gas arrow plot utility so it will be consistent with its own direction keyword, in case that thing ever makes it into IDS so it can be used outside of this test.

### DIFF
--- a/omas/omas_plot.py
+++ b/omas/omas_plot.py
@@ -311,7 +311,7 @@ def gas_arrow(ods, r, z, direction=None, snap_to=numpy.pi/4.0, ax=None, color=No
         """Guesses the direction for the arrow (from injector toward machine) in case you don't know"""
         dr = ods['equilibrium']['time_slice'][0]['global_quantities']['magnetic_axis']['r'] - r
         dz = ods['equilibrium']['time_slice'][0]['global_quantities']['magnetic_axis']['z'] - z
-        theta = -numpy.arctan2(dz, -dr)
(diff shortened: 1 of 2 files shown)


eldond:
@orso82 Ready

orso82:
@eldond just to understand what you have done: you have made the internal calculation in COCOS11, and then modified the output to the `ODS.cocos`. Correct?

eldond:
These are just plots, so their output is only to the screen. Basically, I'm trying to make all the overays tolerate an ODS that's in an arbitrary COCOS. And yes, I'm trying to consistently use COCOS 11 internally.

Gas arrow:
 - The gas arrow thing doesn't actually output anything except the plot command. 
 - I made it treat its internal calculations consistently with its keyword, which I found out it wasn't doing when I wrote a test for its keyword
 - I made it accept the keyword in the COCOS of the ODS by transforming to its internal COCOS

PF coils:
 - Allow for the possibility that the oblique coil angles are stored in arbitrary COCOS by transforming to 11 before plotting
 - All internal calculations are in 11 (with cocos_transform multiplied in as needed)

Issue #28, opened 2018-06-14
Populate omas_cocos.py with missing COCOS transformations


orso82:
@gafusion/omas_developers the OMAS utility 
https://github.com/gafusion/omas/blob/master/utilities/generate_cocos_signals.py
is designed to identify IMAS signals that may require COCOS transforms :)  (Link edited.)

This utility populates the https://github.com/gafusion/omas/blob/master/omas/omas_cocos.py with likely candidates, while retaining the transform information for the entries that already had one.

**Please, as you start using the different IDSs, edit the `omas/utilities/generate_cocos_signals.py` with the missing COCOS transforms.**

This system should also be able to let us keep up with possible future new entries in the IMAS data structure.

eldond:
Is there a way to tell it not to transform that's more explicit than just leaving `'?'` in place?

I tried adding a comment to 
`cocos_signals['pf_active.coil.:.b_field_max']='?'                                                 #[ADD?]# 3.000000 # psi_norm  b_field  [T]`
but it didn't stick. I think this is just a magnitude for material protection, and so it shouldn't get transformed. Right?

orso82:
@eldond @bclyons12 you should now be able to use `cocos_signals['pf_active.coil.:.b_field_max']=None` for entries that do not require transformation. Given these we can the update the logic that identifies the COCOS-relevant quantities here: https://github.com/gafusion/omas/blob/master/omas/omas_physics.py#L328-L392

orso82:
@bclyons12 @eldond I have now tried a few times and I do not see any clobbering of the `omas_cocos.py` file. Maybe you can tell me how to reproduce it?

eldond:
I only tried it when I edited only a comment and nothing else. That is, I left `whatever_thing ='?'` while trying to comment the line.

PR #29, opened 2018-06-14
Update omas_cocos to fill transform types for some hardware positions

eldond:
Mention #28 

When I ran `$ python utilities/generate_cocos_signals.py`, it made a whole bunch of changes to comments, which I didn't commit because it seemed like it would obscure the actual changes I'm proposing. Should I commit the changed comments?


bclyons12:
@eldond @orso82 I also had this problem with generate_cocos_signals.py changing a lot of comments, and in some cases making them incorrect.  See faaa943b3e4fdb3f355e9ecd33adda5b90174bfa

orso82:
@eldond I agree the failed Travis build is because Travis itself is failing to fetch Python 3.5 from S3.

PR #30, opened 2018-06-21
PF active overlay detects geometry type and defines outline path

eldond:
- Plot overlay for pf_active reads geometry type and defines outline path accordingly
  - Supports rectangle and outline so far. Should support arcs of circle at some point, but I don't want to mess with it before I have a real example to work from.
  - Samples and test updated to support new behavior
  - d3d data loader in OMFIT updated by gafusion/OMFIT-source#2836


omas/omas_plot.py
```diff
@@ -845,28 +845,40 @@ def pf_active_overlay(ods, ax=None, **kw):
...
+            [y - dy / 2., y + dy / 2., y + dy / 2., y - dy / 2.]]).T
+
+    def path_outline(outline):
+        """
+        :param outline: ODS sub-folder: element.*.geometry.outline
+        :return: n x 2 array giving the path around the outline of the coil element, suitable for input to Polygon()
+        """
+        return numpy.array([outline['r'], outline['z']]).T
+
+    geo_type_map = ['outline', 'rectangle', 'oblique', 'arcs_of_circle']
```
eldond: @orso82 I defined this in three places (once in OMFIT and twice in OMAS), and that's a problem. I want to define it centrally in OMAS (and import into OMFIT from OMAS) and make it dependent on IMAS version, since I anticipate it changing. What's the best way?

orso82: @eldond define it as a function in `omas.omas_plot.py`, and have OMFIT importing it from omas in `OMFIT-source/classes/omfit_omas.py`. When you do that, place the import in a try statement, in case someone is running an older version of omas. Something like this:

```python
try:
   from omas.omas_plot import path_rectangle, path_outline
except ImportError:
   printe('You are running an old version of OMAS. Please update it.')
```

eldond: @orso82 Okay, I can do that, but shouldn't it be more general than just being in plot? Someone else might need to know how to translate the integer geometry type into the string field name. Maybe it could go in omas_utils?

omas/omas_plot.py
```diff
@@ -363,7 +363,7 @@ def geo_type_lookup(geometry_type, subsys, imas_version=None, reverse=False):
     """
 
     if imas_version is None:
-        imas_version = ODS().imas_version
+        imas_version = '99.99.99'  # Unrecognized
```
orso82: @eldond please always use in your functions

```python
def bla(imas_version=default_imas_version):
   ...
```

eldond: @orso82 YES! Thank you. This is what I was looking for. What's the best way to search OMAS documentation?

orso82: well, it is online: https://gafusion.github.io/omas/code.html

Maybe the question for you is where you would have expected this information to be... we can then put it there ;)

eldond: No, I just needed to realize it was available. Anyway, I think it should be fixed and ready to merge. Let me know if it needs anything else.

PR #40, opened 2018-07-11
Fix bugs in omas_core

bclyons12:


orso82:
Thank you @bclyons12 !

PR #41, opened 2018-07-12
omas_physics updates

bclyons12:
- Update core_physics_pressures
 - Create core_physics_densities


omas/omas_physics.py
```diff
@@ -82,6 +129,82 @@ def core_profiles_pressures(ods, update=True):
...
+
+    for time_index in ods['core_profiles']['profiles_1d']:
+        prof1d = ods['core_profiles']['profiles_1d'][time_index]
+        prof1d_z = ods_z['core_profiles']['profiles_1d'][time_index]
+
+        Z2n = 0.*prof1d_z['grid']['rho_tor_norm']
+        Zn  = 0.*prof1d_z['grid']['rho_tor_norm']
+
+        for k in range(len(prof1d['ion'])):
+            Z = prof1d['ion'][k]['element'][0]['z_n'] # from old ODS
```
orso82: @bclyons12 any reason why you cannot simply work with `ods_z`? The atomic element `z_n` should be there.

bclyons12: Nope, I could do that. I was just following the "read from old, write to new" logic of `core_profiles_pressures` and `core_profiles_densities`

PR #46, opened 2018-08-08
Create transform_current function

bclyons12:


omas/omas_physics.py
```diff
@@ -203,6 +203,47 @@ def core_profiles_zeff(ods, update=True, use_electrons_density=False):
                 prof1d_z['zeff'] = Z2n/Zn
     return ods_z
 
+
+def transform_current(rho, JtoR=None, JparB=None,
```
orso82: Maybe:

```python
def transform_current(equilibrium, rho, JtoR=None, JparB=None, includes_bootstrap=False):
```

or else, I think you should handle better the case where the user does not set `equilibrium`

omas/omas_physics.py
```diff
@@ -203,6 +203,47 @@ def core_profiles_zeff(ods, update=True, use_electrons_density=False):
...
+                      equilibrium=None, includes_bootstrap=False):
+    """
+    Given <Jt/R> returns <J.B>, or vice versa
+    Transformation obeys <J.B> = (1/f)*(<B^2>/<1/R^2>)*(<Jt/R> + dp/dpsi*(1 - f^2*<1/R^2>/<B^2>))
```
orso82: maybe point out in the doc string that both Jtor and JparB must be consistent with the equilibrium.cocosio

omas/omas_physics.py
```diff
@@ -203,6 +203,47 @@ def core_profiles_zeff(ods, update=True, use_electrons_density=False):
...
+    :param rho: normalized rho grid for input JtoR or JparB
+
+    :param JtoR: input <Jt/R> profile (cannot be set with JparB)
+
+    :param JparB: input <J.B> profile (cannot be set with JtoR)
+
+    :param equilibrium: equilibrium ODS containing quanities needed for transformation
+
+    :param includes_bootstrap: set to True if input current includes bootstrap
+
```
orso82: `:return: JtoR if JparB is set, or JparB if JtoR is set

PR #47, opened 2018-08-10
Create update_current function

bclyons12:
Needs thorough testing

### DIFF
--- a/omas/omas_physics.py
+++ b/omas/omas_physics.py
@@ -203,6 +203,131 @@ def core_profiles_zeff(ods, update=True, use_electrons_density=False):
                 prof1d_z['zeff'] = Z2n/Zn
     return ods_z
 
+@add_to__ODS__
+def update_current(ods, time_index, j_ohmic=None, j_bootstrap=None,
+                   j_non_inductive=None, j_total = None, j_tor = None):
+    """
+    This function:
+        - Sets the given currents in ods['core_profiles']['profiles_1d'][time_index]
+        - Updates j_non_inductive, j_total, and/or j_tor if they are not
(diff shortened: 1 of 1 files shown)


omas/omas_physics.py
```diff
@@ -203,6 +203,131 @@ def core_profiles_zeff(ods, update=True, use_electrons_density=False):
...
+            if transform:
+                # transform <J.B>/B0 to <Jt/R>
+                J = transform_current(rho_eq, JparB=J*B0,
+                                      equilibrium=eq, includes_bootstrap=True)
+            else:
+                # already <Jt/R>/<1/R>
+                J *= fsa_invR
+
+            ods.set_time_array('core_profiles.global_quantities.%s'%Iname,time_index,
+                               cumtrapz(vp*J,psi)[-1]/(2.*numpy.pi))
```
orso82: @bclyons12 I may be misunderstanding something in your calculation of the integrated currents, but shouldn't it be like this?

```python
ar=eq['profiles_1d']['darea_dpsi']
...
ods.set_time_array('core_profiles.global_quantities.%s'%Iname,time_index, trapz(ar*J,psi))

```

bclyons12: So `J` here is actually `<Jt/R>`. The `psi` integral turns the flux-surface average into a full volume integral, so you want the volume differential. I've derived this myself before, but don't remember the details off the top of my head. It's the same thing that's done in `OMFITgeqdsk.calc_Ip` though.

bclyons12:
@orso82 It looks like it's all working for me.  Should be good to merge, then I'll update the various STEP and relevant to_omas() functions to use it.

PR #48, opened 2018-08-17
Fix bug in integrated current; j_total can be calculated from j_tor

bclyons12:


smithsp:
@bclyons12 Can you add a check on this additions so that the codecov does not decrease?

orso82:
closing because these changes have been included in #49

PR #49, opened 2018-08-21
Make single function to make core_profiles consistent

bclyons12:


smithsp:
@bclyons12 You should add a test of this function somewhere.

omas/omas_physics.py
```diff
@@ -13,6 +13,31 @@ def add_to__ODS__(f):
...
+
+    :param use_electrons_density:
+            denominator is core_profiles.profiles_1d.:.electrons.density
+            instead of sum Z*n_i in Z_eff calculation
+
+    :return: updated ods
+    '''
+
+    ods_2 = core_profiles_pressures(ods, update=update)
+    ods_2 = core_profiles_densities(ods_2, update=update)
```
orso82: @bclyons12 `update=update` should only be on the first function call. Subsequent functions should all have `update=False`.

omas/omas_physics.py
```diff
@@ -203,6 +228,48 @@ def core_profiles_zeff(ods, update=True, use_electrons_density=False):
                 prof1d_z['zeff'] = Z2n/Zn
     return ods_z
 
+@add_to__ODS__
+def current_from_eq(ods, time_index, set_as='j_ohmic'):
```
orso82: @bclyons12 I am not sure I fully understand the purpose of this function. It looks like you are trying to take the total current from an equilibrium and set it as either the ohmic, bootstrap, or noninductive components in the core_profiles IDS. Shoudn't we set the total current density in the equilibrium IDS equal to the total current density in the core_profiles IDS, and that's it?

omas/omas_physics.py
bclyons12: @orso82 I think you mean that subsequent calls use `update=True`, right?  The call to `core_profiles_pressures` should have `update=update` to make a new ODS if necessary, while the rest operate in place on that.

omas/omas_physics.py
bclyons12: @orso82 Let's discuss in person. The way the `update_current` function works, if you don't set the components of the current, it'll behave strangely.

bclyons12: @orso82 We can merge #48, then I'll update `current_from_eq` here and we can merge that one.

orso82: @bclyons12 I'll merge `update_current` in `core_profiles_consistent` so that it can all be tested here. I'll also work on setting up a regression test for the `update_current()` function with the hope of inspiring you to do the same for `current_from_eq()` \*wink\* \*wink\*  ;)

bclyons12: @orso82 sounds good

orso82:
@bclyons12 please note that I renamed `update_currents` to `core_profiles_currents` in 13347ae . You can probably improve on my basic regression test in 96a287ea1bff36240f3b7ab5461935f720e7db33 and expand it to test `current_from_eq`

PR #51, opened 2018-08-31
Better core_profiles_currents

bclyons12:


bclyons12:
@orso82 f54f28a provides an extensive regression test for `core_profiles_currents`.  I'm still getting "All checks have failed" here, but `make tests` seems to run fine for me.  Not sure if those two things are even related.

orso82:
image

orso82:
It looks like in the `numpy.allclose` routine you are passing two elements that are not of the same type (one is a string (or unicode, because it's python3) and the other is a numpy array).

PR #52, opened 2018-09-18
Edit docstring format to avoid sphinx warnings

eldond:
- Built docs in OMFIT. See gafusion/OMFIT-source#3129

### DIFF
--- a/omas/omas_itm.py
+++ b/omas/omas_itm.py
@@ -261,8 +261,7 @@ def save_omas_itm(ods, user=None, machine=None, shot=None, run=None, new=False, 
     :param new: whether the open should create a new ITM tree
 
     :param itm_version: ITM version
-                         (reads ods['info.itm_version'] if itm_version is None
-                          and finally fallsback on itm version of current system)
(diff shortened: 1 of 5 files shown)


eldond:
In addition to checking for clearing warnings from `sphinx.warnings`, I also opened `code.html` (also in OMFIT) and looked for weird formatting issues, which correspond with the warnings.

omas/omas_itm.py
```diff
@@ -355,7 +354,7 @@ def save_omas_itm(ods, user=None, machine=None, shot=None, run=None, new=False,
 # AUTOMATICALLY GENERATED FILE - DO NOT EDIT
 
 def load_omas_itm(user=os.environ['USER'], machine=None, shot=None, run=0, paths=None,
-                   itm_version=default_itm_version, verbose=None):
```
orso82: @eldond the `omas_itm.py` file is automatically generated from `omas_imas.py`.
You can automatically generate it by running `make itm` in the main omas repository.

PR #64, opened 2019-04-01
 os.environ['USER'] -> os.environ.get('USER','dummy_user')

kripnerl:
In some environment (for example for CI). USER is not specified.


orso82:
handled via bd11345
Thank you @kripnerl !

PR #72, opened 2019-07-15
Bugfix/dynamic path creation

Karel-van-de-Plassche:
`self.dynamic_path_creation` is a boolean, not a string. Looks like this blob of code is not tested, I'm not sure when it's supposed to trigger, so please add a test.


orso82:
addressed with 13cf16003281e09b00ac6df30b6baaca6c0b54a6

PR #73, opened 2019-07-15
Added setting deeper array index

Karel-van-de-Plassche:
Seeing from the lines above, this would be expected behaviour. Now it triggers also when setting 'regular' deep arrays indices. If this is not expected, please add a test that tests the contrary. Fixed typo in index check error.


orso82:
Thank you @Karel-van-de-Plassche 👍  I ended up somewhat refactoring that part to avoid code repetition and improve readability/logic.

Karel-van-de-Plassche:
Great, look good!

PR #83, opened 2020-02-06
Added stored energy calculation in equilibrium ods called in equilibrium_consistent()

TimSlendebroek:


orso82:
Great! Thank you @TimSlendebroek 

Could you please add it to the regression tests under `test_omas_physics` https://github.com/gafusion/omas/blob/TSdev_branch/omas/tests/test_omas_physics.py

orso82:
@TimSlendebroek you are right that `equilibrium_consistent()` would test your script, but (my mistake) we do not test that. In fact, if you could add a `test_equilibrium_consistent()` routine that would be great!

TimSlendebroek:
Sure thing, I will look around the rest of the tests what the format is. I guess this will show up in the automatic codecov that is being done on the codebase?

orso82:
Yes. It should show up on the codedov and hopefully not as a failed test on the TravisCI regressions :) NOTE: You can run `make tests` in the top level `omas` folder to run the same regressions that TravisCI would run.

TimSlendebroek:
added reg test!

orso82:
It looks like you also found a bug in doing so ;) 👍

omas/omas_physics.py
```diff
@@ -29,6 +29,29 @@ def add_to__ALL__(f):
...
+    '''
+    ods_n = ods
+    if not update:
+        from omas import ODS
+        ods_n = ODS().copy_attrs_from(ods)
+
+    for time_index in ods['equilibrium']['time_slice']:
+        pressure_equil = ods['equilibrium']['time_slice'][time_index]['profiles_1d']['pressure']
+        volume_equil = ods['equilibrium']['time_slice'][time_index]['profiles_1d']['volume']
+        dvol = numpy.gradient(volume_equil)
```
smithsp: Why calculate `dvol`?  Why not just give `volume_equil` to `numpy.trapz` as the keyword `x`?  (see https://numpy.org/doc/1.18/reference/generated/numpy.trapz.html#r7aa6c77779c0-2)

PR #86, opened 2020-02-11
Fix up options for plotting X-points in plot overlays

eldond:
- Add missing item to docstring
- Add control of measured X-point plot marker in shape target overlay
  - old efitviewer includes measured X-point here, so that's why it's there
- Add measured X-point as an option in equilibrium CX
image


omas/version
```diff
@@ -1 +1 @@
-0.52.1
+0.52.2
```
orso82: @eldond changing the version number requires a whole series of actions (git tagging, release to pypi, update docs). Please leave that to me, so that I can do that.

eldond: Okay. I am tying OMFIT developments to OMAS versions, though, so I need some input to request a version increment in order to continue doing that. That is, I have features which only activate in OMFIT if OMAS version is >= 0.52.2. If I shouldn't increment the OMAS version number, should I instead inspect relevant function docstrings for keywords that would indicate the relevant features are present?

orso82: I would discourage you to do that. You are going to carry a lot of extra code around to do that. Such effort is really only necessary if you have a lot of users and do not want to break backward compatibility for them. If it's mostly you working on the code right now, just make things work with the latest version...

eldond: morti can't pass the tests with the old OMAS versions unless I put these backwards compatibility checks in. Installations on iris, etc. will have older OMAS. I can't just have the GUI not open because someone is on OMAS 0.52.0 instead of 0.52.1. Besides, I defined one dictionary of keywords and required OMAS versions. It goes through and pops out keywords to OMAS calls if the version requirements aren't met, so it's actually pretty low maintenance, if I can tie everything to an OMAS version number. I could define a list of supported features in OMFIT instead.

orso82: Your choice. You can then increase the version number. There will just not be a PIP version for these releases that you are making.

eldond: Is that acceptable if I only change the last field? If I update `0.52.x`, `0.52.y`, ... and you officially release `0.52.0` and `0.53.0`, all my logic will work fine. If that's okay with you, I think that would be smoothest for me. I don't want to make a mess, though.

orso82: Yes. It's not big deal if you increment the last digit :)

PR #79, opened 2019-11-27
IMAS <> GKDB compatibility

Karel-van-de-Plassche:
I've merged a small modification of the GKDB.json into master. This updates the .json file you copied accordingly.

**Some background how this happend**
GKDB data format was originally explicitly not IMAS-compatible, as we wanted to be more flexible, as well have a structure that can be easily stored in noSQL/SQL databases. As such, the GKDB JSON specification was born. After it was 'finished', IMAS took the structure and adopted it as their 'gyrokinetic IDS'. Unfortunately this adoption was not 1-to-1, and as such the GKDB JSON was not IMAS compatible anymore (it was never intended as such)

**What I've done**
As the GKDB JSON is so extremely close to OMAS-style IMAS compatibility, I've just changed the GKDB JSON specification. Now it's IMAS compatible, but not OMAS-IMAS compatible yet. I'm adding OMAS compatibility tests in here: https://gitlab.com/gkdb/gkdb/merge_requests/2, combined with a directory cleanup. Feel free to join in.

**What needs to be done still** (In this pull request or in a separate one):
Looks like the data shapes are not correctly handled in the pythonization of the JSON. Specifically two issues that fail the coordinate shape test here [omas_core.py#L622](https://github.com/gafusion/omas/blob/master/omas/omas_core.py#L622), which I don't think should fail:
- For eigenvalue runs, `time_norm` has no meaning. In the GKDB JSON format, we allow a field being `null` (or `None` in Python). This should be interpreted as that the coordinate has size 1. If you think the field should be `[null]` instead of `null` for easier integration, we can change that
- `phi_potential_perturbed_norm_real` and related fields can be 1D-ish, as the timebase can have length one. Seems like OMAS instead squeezes out this length-1 dimension, resulting in a wrongly shaped array (1D instead of 2D), thus failing the test
- Looks like the load instance methods `ODS.load` and `ODS.load_structure` disappeared, so the example currently fails

**What I hope to do in the very long term**
Open source the IMAS-DD and associated tools. Discussion ongoing, but slow, as it is subject to the ITER rules and legislation.

PR #87, opened 2020-02-11
Add all `examples` as regression tests

orso82:

PR #88, opened 2020-02-12
Updated GKDB JSON to latest version; now IMAS compatible

Karel-van-de-Plassche:
We updated the GKDB JSON to be more compatible with IMAS/OMAS. Continuation of #79, see there for a longer explanation.


orso82:
Super! Also, with #87 I am adding all of the examples scripts to be run as part of the regression tests!

PR #91, opened 2020-02-24
Improvements to CX overlays

eldond:
- Bugfix: `equilibrium_CX()` was overriding its levels keyword
- Allow customization of pos ctrl point labels


eldond:
@orso82 it's ready

PR #92, opened 2020-03-04
added try except to make sure a rho or psi grid is initialized in core_profiles_pressures

TimSlendebroek:


omas/omas_physics.py
```diff
@@ -117,7 +118,10 @@ def core_profiles_pressures(ods, update=True):
         if not update:
             prof1d_p['grid']['rho_tor_norm'] = prof1d['grid']['rho_tor_norm']
 
-        __zeros__ = 0. * prof1d['grid']['rho_tor_norm']
```
orso82: The `core_profiles` ODS should always have `rho_tor_norm` defined, since that's the coordinate for all 1d profiles

smithsp: @TimSlendebroek There were two of these instances of `try`

TimSlendebroek: @smithsp I think I deleted them both now and only kept the np.trapz change

PR #94, opened 2020-03-10
More robust position_control overlay and stricter warnings during tests

eldond:
- position_control overlays:
  - Don't fail the whole overlay if just one out of boundary, strike, or X-point is missing
  - Fail if they're all missing
  - Allows plotting of boundary even if X-point and strike point are missing
- testing:
  - Was experiencing a bad interaction between OMAS and OMFIT imports. Upgraded the warnings to try to catch it.

### DIFF
--- a/omas/tests/failed_imports.py
+++ b/omas/tests/failed_imports.py
@@ -1,7 +1,7 @@
 from omas.omas_setup import omas_rcparams
 import os
 import warnings
-from omas.tests.warning_setup import hard_warnings, set_omas_warnings
+from .tests.warning_setup import hard_warnings, set_omas_warnings
 
 try:
     import imas
--- a/omas/tests/failed_imports.py
+++ b/omas/tests/failed_imports.py
@@ -46,15 +46,13 @@ try:
 except (ImportError, ServerSelectionTimeoutError) as _excp:
     failed_MONGO = _excp
 
-if hard_warnings:
+with warnings.catch_warnings():
     warnings.simplefilter('ignore')
-try:
-    from omfit.classes.omfit_eqdsk import OMFITgeqdsk, OMFITsrc
+    try:
+        from omfit.classes.omfit_eqdsk import OMFITgeqdsk, OMFITsrc
 
-    failed_OMFIT = False
-except ImportError as _excp:
-    failed_OMFIT = _excp
-if hard_warnings:
-    set_omas_warnings()
+        failed_OMFIT = False
+    except ImportError as _excp:
+        failed_OMFIT = _excp
 
 __all__ = ['failed_IMAS', 'failed_HDC', 'failed_S3', 'failed_MONGO', 'failed_OMFIT', 'failed_UDA']
--- a/omas/tests/failed_imports.py
+++ b/omas/tests/failed_imports.py
@@ -1,7 +1,7 @@
 from omas.omas_setup import omas_rcparams
 import os
 import warnings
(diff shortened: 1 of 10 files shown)


omas/tests/test_omas_utils.py
```diff
@@ -64,10 +65,12 @@ def test_different_ods(self):
         diff_prof5 = different_ods(ods2, ods3)
         assert isinstance(diff_prof5, list)
         assert 'name' in ' '.join(diff_prof5)
+        return
```
orso82: @eldond is there a reason for adding `return` at the end of each function. That's not necessary, since the function will also return `None`

omas/tests/test_omas_utils.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
orso82: python3 and not python

omas/tests/failed_imports.py
```diff
@@ -44,11 +46,15 @@
 except (ImportError, ServerSelectionTimeoutError) as _excp:
     failed_MONGO = _excp
 
+if hard_warnings:
```
orso82: Best done with a `with` statement, starting at the beginning of this file

omas/tests/test_omas_utils.py
eldond: It's easier to read and easier to fold

omas/tests/test_omas_utils.py
eldond: what do you mean?

omas/tests/failed_imports.py
eldond: Yes, good point

omas/tests/test_omas_utils.py
eldond: nevermind, I get it. Fixed.

omas/tests/test_omas_utils.py
```diff
@@ -118,22 +125,28 @@ def test_omas_info(self):
...
         ods.set_time_array('equilibrium.vacuum_toroidal_field.b0', 0, 0.1)
         ods.set_time_array('equilibrium.vacuum_toroidal_field.b0', 1, 0.2)
+        return
 
     def test_info(self):
         omas_info('equilibrium')
         omas_info(None)
+        return
+
+    # End of TestOmasUtils class
```
eldond: This also makes the class end easier to find, and makes it easier to fold (at least for me in pycharm) the whole class since the last line of the class is no longer the last line of a function.

omas/tests/test_omas_utils.py
eldond: To clarify, I've found that some of these classes with a ton of methods get messy without nice clear breaks between methods. So, I like to put in the empty returns. It's just a preference/style thing. Go ahead and delete them if you don't like them.

orso82: Yeah, I do not like them, and I would not want to add it to every function/method to be consistent. For now I can survive with these empty return statements and perhaps I'll remove them later.

I am not sure I understand what you mean by `easier to fold` in PyCharm. You'll have to show me!

eldond: Sometimes the last line of the function is also the last line in an indented block for if, else, for, or something. Then it seems to default to folding the if/else/for block instead of the whole function. The same thing for classes, where the last line of the class is the last line of a method, unless a terminating comment or something is added.

PR #95, opened 2020-03-23
Fix missing 'data_type' in consistency_checker

bclyons12:
This was causing an error when trying to enter a location in OMFIT using `consistency_check='warn'`


eldond:
@bclyons12 Your build is failing because of the numpy version that travis is using. The requirements file doesn't specify a version, but if you specify one, it can pass. I fixed this in #96.

bclyons12:
@eldond I don't understand. I hadn't changed any files or requirements besides editing one line.

eldond:
@bclyons12 As part of the test process, travis sets up a fresh virtual machine from a template and does `pip install .` to install the repo. If new versions of packages come out, the installed version can change, unless the requirement file specifies specific versions. The OMAS requirements were not specific to any numpy version, so our build was vulnerable to changes in our dependencies.

PR #118, opened 2020-07-06
Fix problems with labeling gas injectors

eldond:
- Gas injection direction fix
    - Gas injectors have an exit location and also a second_point
    - Together, they define the direction of injection, where the exit is the source and the second_point is the target (towards the plasma)
    - omas_plot bundles pipes that are at the same poloidal location and uses one arrow to represent the group, based on the second point location (r2, z2) of the last injector of the group 
    - Now, if the last pipe in the group doesn't have r2, z2, it will try to copy r2, z2 from the previous pipe.
- Labels for groups of injectors
    - Option to print the names of all the injectors at a particular R,Z exit location (on different lines)

image

### DIFF
--- a/omas/omas_plot.py
+++ b/omas/omas_plot.py
@@ -2005,7 +2005,7 @@ def gas_injection_overlay(
     for i, loc in enumerate(locations):
         r, z = numpy.array(loc.split('_')).astype(float)
         if show_all_pipes_in_group:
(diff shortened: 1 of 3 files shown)


orso82:
@eldond let me know when it is ready to merge

eldond:
@orso82 Ready when it passes all the checks.

eldond:
@orso82 It's ready.

PR #119, opened 2020-07-25
added physics_total_powers method to ods and couple it to tau_e

TimSlendebroek:
new method for ods.physics_summary_total_powers, should throw errors if the source.identifier.index isn't filled!

Should be easy to add systems or totals if they are added to the data schema by appending the two dicts.


TimSlendebroek:
I have a feeling that the regression test failing isn't my fault this time

jmcclena:
👍

omas/omas_physics.py
```diff
@@ -267,9 +267,15 @@ def summary_total_powers(ods, update=True):
...
               'ic': copy.deepcopy(q_init)}
 
+    ignore_indices = list(range(100, 108)) + list(range(900,910))
+
     for time_index in sources[0]['profiles_1d']:
         vol = sources[0]['profiles_1d'][time_index]['grid']['volume']
         for source in sources:
+            if sources[source]['identifier.index'] in ignore_indices:
+                # Skip the combined sources to prevent double counting
+                print(sources[source]['identifier.name'], "is excluded from summary_total_powers with index", sources[source]['identifier.index'])
```
orso82, line 277: If you know that's the right thing to do, I'd be ok not to `print` anything on screen.

omas/omas_physics.py
```diff
@@ -249,6 +242,63 @@ def summary_taue(ods, update=True):
...
+    :param update: operate in place
+
+    :return: updated ods
+    """
+    ods_n = ods
+    if not update:
+        from omas import ODS
+        ods_n = ODS().copy_attrs_from(ods)
+
+    import copy
```
orso82, line 261:
```suggestion
```

omas/omas_physics.py
```diff
@@ -249,6 +242,63 @@ def summary_taue(ods, update=True):
...
+        vol = sources[0]['profiles_1d'][time_index]['grid']['volume']
+        for source in sources:
+            if sources[source]['identifier.index'] in ignore_indices:
+                # Skip the combined sources to prevent double counting
+                print(sources[source]['identifier.name'], "is excluded from summary_total_powers with index", sources[source]['identifier.index'])
+                continue
+            if 'electrons' in sources[source]['profiles_1d'][time_index] and 'energy' in \
+                    sources[source]['profiles_1d'][time_index]['electrons']:
+                q_dict['total'] += sources[source]['profiles_1d'][time_index]['electrons']['energy']
+                if sources[source]['identifier.index'] in index_dict.keys():
```
orso82, line 282:
```suggestion
                if sources[source]['identifier.index'] in index_dict:
```

omas/omas_physics.py
```diff
@@ -249,6 +242,63 @@ def summary_taue(ods, update=True):
...
+                    q_dict[index_dict[sources[source]['identifier.index']]] += \
+                    sources[source]['profiles_1d'][time_index]['electrons']['energy']
+            if 'total_ion_energy' in sources[source]['profiles_1d'][time_index]:
+                q_dict['total'] += sources[source]['profiles_1d'][time_index]['total_ion_energy']
+                if sources[source]['identifier.index'] in index_dict.keys():
```
orso82, line 287:
```suggestion
                if sources[source]['identifier.index'] in index_dict:
```

omas/omas_physics.py
TimSlendebroek: @orso82  Right now nobody is using the combined sources, 101 - 109 but I am worried that in the future these might be populated instead of the individual ones (Ic & ec combined), this this also excludes self-made indices 901-909. 

But I agree it would be annoying if it is being printed to the screen so I will delete it 👍

omas/omas_physics.py
TimSlendebroek: makes sense :)

omas/omas_physics.py
TimSlendebroek: makes sense :)

PR #121, opened 2020-09-09
Minor fixes

TimSlendebroek:
No need to sum density (density_thermal + density_fast = density) double counting

Not all ODSses have psi in core_profiles so added that in the if statement


omas/omas_plot.py
```diff
@@ -1186,7 +1186,7 @@ def core_transport_fluxes(ods, time_index=None, fig=None, show_total_density=Tru
 
     def sum_density_types(specie_index):
         final_density = numpy.zeros(len(prof1d['grid.rho_tor_norm']))
-        for therm_fast in ['', '_thermal', '_fast']:
+        for therm_fast in ['_thermal', '_fast']:
```
orso82, line 1189: @TimSlendebroek I'd run `ods.physics_core_profiles_densities()` to make sure that `density`, `density_thermal`, and `density_fast` for electrons and ions are filled and are self-consistent.

PR #126, opened 2020-09-22
Add start of list of formatted physics symbols for IMAS quantities

eldond:
- Goal: make it easy to get nicely formatted physics symbols for quantities

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,8 +62,11 @@ json:
 	cd omas/utilities && python3 build_json_structures.py
 	make cocos
 
-cocos:
+cocos: symbols
 	cd omas/utilities && python3 generate_cocos_signals.py
+
+symbols:
+	cd omas/utilities && python3 sort_symbols.py
 
 tag:
 	git tag -a v$(VERSION) $$(git log --pretty=format:"%h" --grep="^version $(VERSION)") -m "version $(VERSION)"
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -62,9 +62,12 @@ json:
 	cd omas/utilities && python3 build_json_structures.py
 	make cocos
 
-cocos:
+cocos: symbols
```
eldond, line 65: @orso82 Is this right? Seems like the next entry might've bled through here by accident, but I'm not sure.

eldond:
@orso82 Okay, great. This makes it pretty clear what needs to be done to extend the symbols later. Then to use the symbols, I just import PhysicsSymbols, right? OOHHH and you made it so I can just feed in whatever location I want without bothering to strip/split/whatever it myself, didn't you!? So convenient!

omas/tests/test_omas_core.py
```diff
@@ -427,6 +427,20 @@ def test_codeparameters(self):
         assert len(ods['ec_launchers.code.parameters.launcher.0']) == 2
         return
 
+    def test_latexit(self):
+        assert latexit['somewhere.:.sublocation.n_e'] == '$n_e$'
+        assert latexit['n_e'] == '$n_e$'
+        assert latexit['.n_e'] == '$n_e$'
+        assert latexit['n_e'] == '$n_e$'
```
smithsp, line 434: What is the difference between this and two lines prior?

eldond:
@orso82 I'm finding it pretty convenient that you rigged it so I can do `latexit[thing]` instead of `latexit(thing)`, because I can also do `latexit.get(thing, thing)` and that's really clear.

Issue #125, opened 2020-09-21
Length of lists/arrays of strings accessed by `.:.` is set by first element and not extended later, which causes strings to be truncated


eldond:
Sample code:
``` python
OMFIT['ods'] = ods = ODS()
ods['langmuir_probes.embedded.0.name'] = 1
ods['langmuir_probes.embedded.1.name'] = 10
ods['langmuir_probes.embedded.2.name'] = 100
print(ods['langmuir_probes']['embedded.:.name'])
```
Actual result:
```
['1' '1' '1']
```
Desired result:
```
['1' '10' '100']
```
These strings happen to contain numbers this time because that's how KSTAR probes are designated. But the general case is for these to be strings. It works as desired if probe 0 has the longer number, but... uh... that's not how numbers usually work.

eldond:
I bet this is related to #124

Issue #124, opened 2020-09-21
Trouble accessing `langmuir_probes.embedded.:.time`


eldond:
Langmuir probes are allowed to have independent timing for different probes. That is, they use `langmuir_probes.embedded.:.time`, not `langmuir_probes.embedded.time`.
Accessing `langmuir_probes.embedded.:.time` SOMETIMES works, and it seems that some padding with NaNs can be done to account for missing times. However, it doesn't work for varying array lengths, which are allowed in IMAS and make sense on the basis of how LPs work. It would be useful if `times = ods['langmuir_probes.embedded.:.time']` would return a NaN padded 2D array for an ODS with varying lengths of time arrays, uneven time-bases, and missing time information for some probes (it's normal and useful to record probe physical R-Z coordinates even when they're off and have no data or time).

Here are commands for filling in sample data in an ODS that won't work with `times = ods['langmuir_probes.embedded.:.time']`.
``` python
ods['langmuir_probes.embedded.0.name'] = '1'
ods['langmuir_probes.embedded.0.position.r'] = 1.235
ods['langmuir_probes.embedded.0.position.z'] = -0.952
ods['langmuir_probes.embedded.0.position.phi'] = 0.0
ods['langmuir_probes.embedded.1.name'] = '2'
ods['langmuir_probes.embedded.1.position.r'] = 1.3
ods['langmuir_probes.embedded.1.position.z'] = -0.99
ods['langmuir_probes.embedded.1.position.phi'] = 0.0
ods['langmuir_probes.embedded.2.name'] = '3'
ods['langmuir_probes.embedded.2.position.r'] = 1.332
ods['langmuir_probes.embedded.2.position.z'] = -1.023
ods['langmuir_probes.embedded.2.position.phi'] = 0.0
ods['langmuir_probes.embedded.3.name'] = '4'
ods['langmuir_probes.embedded.3.position.r'] = 1.36
ods['langmuir_probes.embedded.3.position.z'] = -1.0295
ods['langmuir_probes.embedded.3.position.phi'] = 0.0
ods['langmuir_probes.embedded.4.name'] = '5'
ods['langmuir_probes.embedded.4.position.r'] = 1.381
ods['langmuir_probes.embedded.4.position.z'] = -1.032
ods['langmuir_probes.embedded.4.position.phi'] = 0.0
ods['langmuir_probes.embedded.5.name'] = '6'
ods['langmuir_probes.embedded.5.position.r'] = 1.41
[...]

PR #127, opened 2020-09-23
Data slice fixes

orso82:


omas/tests/test_omas_core.py
```diff
@@ -127,6 +127,22 @@ def test_misc(self):
...
+    def test_data_slicing(self):
+        ods = ODS()
+        ods['langmuir_probes.embedded.0.name'] = '1'
+        ods['langmuir_probes.embedded.1.name'] = '12'
+        ods['langmuir_probes.embedded.2.name'] = '123'
+        assert ods['langmuir_probes']['embedded.:.name'][2] == '123'
+
+        ods = ODS(consistency_check=False)
+        for k in range(3):
+            ods[f'langmuir_probes.embedded.{k}.name'] = [float(k)] * (k + 1)
```
eldond, line 139: Doesn't assigning this into 'name' force it to be a string? Maybe put it into ...embedded.{k}.time or ...embedded.{k}.saturation_current_ion.data or something instead? I can't even make this work in the OMFIT command box because it wants a string (`ValueError: langmuir_probes.embedded.1.name is of type <class 'numpy.ndarray'> must be a scalar of type STR_0D`).

omas/tests/test_omas_core.py
```diff
@@ -127,6 +127,22 @@ def test_misc(self):
         )
         return
 
+    def test_data_slicing(self):
+        ods = ODS()
+        ods['langmuir_probes.embedded.0.name'] = '1'
+        ods['langmuir_probes.embedded.1.name'] = '12'
+        ods['langmuir_probes.embedded.2.name'] = '123'
+        assert ods['langmuir_probes']['embedded.:.name'][2] == '123'
```
eldond, line 135: This is the right test for #125 and that issue can be closed if this passes

eldond (changes requested):
My manual testing suggests that this should close both #125 and #124, but the test concerning #124 is wrong; it tries to assign arrays into `name`, which is a string field, and I'm not sure how it could've ever passed. Just changing name to time in the later part of the test unit should fix it, though.

omas/tests/test_omas_core.py
orso82: @eldond notice that I disabled the IMAS consistency checks with `consistency_check=False`

eldond: Okay. I should've seen that.

eldond (approved):
It's fine, but the test would be more clear (maybe someone will look at it as an example?) if you change name to time for the array part.

PR #128, opened 2020-09-25
OMAS Data Collection class

orso82:
The ODC class allows collating data from multiple ODSs into a single object.
ODC supports the same functionalities as ODSs.

### DIFF
--- a/omas/examples/omas_collection.py
+++ b/omas/examples/omas_collection.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+"""
+OMAS data collection
+=====================
+ODC class can be used to collate ODSs, which is useful for collection of data from:
+* different experimental shots
+* different simulations
+"""
+
+from omas import *
+
+print('# collection identifier as a string')
+odc = ODC()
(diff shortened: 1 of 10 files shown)


orso82:
@eldond and @jmcclena note that ODCs can only be saved/loaded as pickle files at this point

eldond (approved):
Wow, good job. It just works, and it's smooth. I tried this:
``` python
ods1 = OMFIT['ods1'] = ODS()
multi_efit_to_omas('DIII-D', 180257, 'EFIT01', ods=ods1)
ods2 = OMFIT['ods2'] = ODS()
multi_efit_to_omas('DIII-D', 180264, 'EFIT01', ods=ods2)
odc = OMFIT['odc'] = ODC()
odc['aaa1'] = ods1
odc['bbb1'] = ods2
print(odc[':.equilibrium.time'])
print(shape(odc[':.equilibrium.time_slice.:.global_quantities.psi_axis']))
print(shape(odc[':.equilibrium.time_slice.:.profiles_2d.0.grid.dim1']))
print(shape(odc[':.equilibrium.time_slice.:.profiles_2d.0.psi']))
```
and it just makes sense. Is there anything else specific I should try? I feel like I already know how it works, since it's acting like an ODS that doesn't have restrictions on the first level of the hierarchy.

PR #130, opened 2020-10-06
add line average density calculation to omas

jmcclena:
Hasn't been tested on omas! Don't merge yet.

@orso82 @TimSlendebroek here is my idea for calculating the line-average density in omas.

### DIFF
--- a/omas/omas_physics.py
+++ b/omas/omas_physics.py
@@ -248,7 +248,7 @@ def summary_lineaverage_density(ods, line_grid=2000, time_index=None, update=Tru
             x2 = R2*numpy.cos(phi2)
             y2 = R2*numpy.sin(phi2)
                 
-            xline = (x2-x1) * numpy.linspace(0, 1, line_grid) + x1
+            xline = numpy.linspace(x1, x2, line_grid)
             yline = (y2-y1) * numpy.linspace(0, 1, line_grid) + y1			
             Rline = (R2-R1) * numpy.linspace(0, 1, line_grid) + R1
             Zline = (Z2-Z1) * numpy.linspace(0, 1, line_grid) + Z1
@@ -261,11 +261,9 @@ def summary_lineaverage_density(ods, line_grid=2000, time_index=None, update=Tru
             i1 = zero_crossings[0]
             i2 = zero_crossings[-1]
 
-            ne_line = 0.
-            for i in range(i1, i2, numpy.sign(i2-i1)):
-                psival = psi_interp(Zline[i], Rline[i])[0]
-                ne_interp  = scipy.interpolate.splev(psival, tck)
-                ne_line  += ne_interp
+            psival = [psi_interp(Zline[i], Rline[i])[0] for i in range(i1, i2, numpy.sign(i2-i1))]
+            ne_interp  = scipy.interpolate.splev(psival, tck)
+            ne_line  = numpy.trapz(ne_interp)
             ne_line /= abs(i2-i1)
             ne_line_paths.append(ne_line)
             dist_paths.append(numpy.sqrt((xline[i2]-xline[i1])**2+(yline[i2]-yline[i1])**2+(Zline[i2]-Zline[i1])**2))
--- a/omas/omas_physics.py
+++ b/omas/omas_physics.py
@@ -94,6 +94,57 @@ def equilibrium_stored_energy(ods, update=True):
 
 
 @add_to__ODS__
+def calc_line_den(ods, R1, R2, Z1, Z2, line_grid=2000, time_index=0):
+    ods_n = ods
+    if not update:
+        from omas import ODS
+
+        ods_n = ODS().copy_attrs_from(ods)
+
+    Rb = ods['equilibrium']['time_slice'][time_index]['boundary']['outline']['r']
+    Zb = ods['equilibrium']['time_slice'][time_index]['boundary']['outline']['z']
+
+    Rline = (R2-R1) * numpy.linspace(0, 1, line_grid) + R1
+    Zline = (Z2-Z1) * numpy.linspace(0, 1, line_grid) + Z1
+    dist = numpy.zeros(line_grid)
+
+    for i, Rval in enumerate(Rline):
+        dist[i]= numpy.min((Rline[i]-Rb)**2 + (Zline[i]-Zb)**2 )
+
+    a = numpy.argpartition(dist, 4)[0: 3]
+    i1 =  a[0]
+    i2 = a[1]
+    if abs(i2-i1) == -1:
+        i2 = a[3]
+
+    Rgrid = ods['equilibrium']['time_slice'][time_index]['profiles_2d'][0]['grid']['dim1']
+    Zgrid = ods['equilibrium']['time_slice'][time_index]['profiles_2d'][0]['grid']['dim2']
+
+    psi2d = ods['equilibrium']['time_slice'][time_index]['profiles_2d'][0]['psi']
+    psi_interp = scipy.interpolate.interp2d(Rgrid, Zgrid, psi2d)
+    psi_eq = ods['equilibrium']['time_slice'][time_index]['profiles_1d']['psi']
+    rhon_eq = ods['equilibrium']['time_slice'][time_index]['profiles_1d']['rho_tor_norm']
+    rhon_cp = ods['core_profiles']['profiles_1d'][time_index]['grid']['rho_tor_norm']
+    ne = ods['core_profiles']['profiles_1d'][0]['electrons']['density_thermal']
+    ne = numpy.interp(rhon_eq, rhon_cp, ne)
+    tck = scipy.interpolate.splrep(psi_eq, ne, k=3)
+    ne_line = 0.
+    for i in range(i1, i2, numpy.sign(i2-i1)):
+        psival = psi_interp(Rline[i], Zline[i])[0]
+        ne_interp  = scipy.interpolate.splev(psival, tck)
+        ne_line  += ne_interp
+    ne_line /= abs(i2-i1)*sqrt((Rline[i2]-Rline[i1])**2+(Zline[i2]-Zline[i1])**2)
+
+
+    if 'summary.line_average.n_e' in ods:
+        ods_n['summary.line_average.n_e'][time_index] = ne_line
+    else:
+        ods_n['summary.line_average.n_e'][time_index] = numpy.atleast1d(ne_line)
+        
+    return ne_line
+
+
+@add_to__ODS__
 @preprocess_ods('equilibrium')
 def equilibrium_ggd_to_rectangular(ods, time_index=None, resolution=None, method='linear', update=True):
     """
@@ -1738,7 +1789,7 @@ def generate_cocos_signals(structures=[], threshold=0, write=True, verbose=False
 #        transforms['BP'] = transforms['POL']
 #        transforms[None] = 1
 #
-                
+
 _cocos_signals = {}
 """
         ]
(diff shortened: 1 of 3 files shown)


TimSlendebroek:
nice addition, I was trying to add this to tau_e_h98 and Greenwald. What are you giving to the following params:
 R1, R2, Z1, Z2 in the function def calc_line_den(ods, R1, R2, Z1, Z2) @jmcclena

omas/omas_physics.py
```diff
@@ -93,6 +93,58 @@ def equilibrium_stored_energy(ods, update=True):
...
 
+@add_to__ODS__
+@preprocess_ods('core_profiles', 'equilibrium')
+def calc_line_den(ods, R1, R2, Z1, Z2, line_grid=2000, time_index=0):
+    ods_n = ods
+    if not update:
+        from omas import ODS
+
+        ods_n = ODS().copy_attrs_from(ods)
+
```
orso82, line 104: @jmcclena can you make it so that by default `time_index=None` and if so it calculates it for all time slices?

You can simply call your function recursively:

```suggestion
    if time_index is None:
        for time_index in range(len(ods['equilibrium']['time_slice'])):
            calc_line_den(ods_n, R1=R1, R2=R2, Z1=Z1, Z2=Z2, line_grid=line_grid, time_index=time_index, update=False)
        return ods_n
```
```

Comment (jmcclena) [reply omas/omas_physics.py:104]: > nice addition, I was trying to add this to tau_e_h98 and Greenwald. What are you giving to the following params:
> R1, R2, Z1, Z2 in the function def calc_line_den(ods, R1, R2, Z1, Z2) @jmcclena
@TimSlendebroek 
This function should work for any two points that pass through the plasma boundary. For DIII-D chords, it should be.

V1 chord: R1 =  R2 = 1.48 m, Z1=1.5, Z2=-1.5
V2 chord: R1 =  R2 = 1.94 m, Z1=1.5, Z2=-1.5
V3 chord: R1 =  R2 = 2.10 m, Z1=1.5, Z2=-1.5
R chord: Z1=Z2=0, R1 = 1.0 and R2 = 2.4

Comment (orso82) [reply omas/omas_physics.py]: @jmcclena you can add a DIII-D specific function in OMFIT under `OMFIT-source/omfit/classes/omfit_omas_d3d.py`
@eldond may be able to suggest how to best do this.

Also, could you also add a regression test that covers this new physics function?

File: omas/omas_physics.py [hunk 2]
```diff
@@ -180,6 +181,74 @@ def summary_greenwald(ods, update=True):
     ods_n['summary.global_quantities.greenwald_fraction.value'] = abs(numpy.array(ne_vol_avg) / 1e20 / ip * 1e6 * numpy.pi * a ** 2)
     return ods_n
 
+@add_to__ODS__
+@preprocess_ods('core_profiles', 'equilibrium')
+def summary_lineaverage_density(ods, R1, R2, Z1, Z2, line_grid=2000, time_index=None, update=True):
```
eldond, line 186: The default behavior should be to read interferometer geometry from the ODS. It's in:
`interferometer.channel[:].line_of_sight.first_point.r`
`interferometer.channel[:].line_of_sight.first_point.z`
`interferometer.channel[:].line_of_sight.first_point.phi`
`interferometer.channel[:].line_of_sight.second.r`
...
`interferometer.channel[:].line_of_sight.third_point.r`
...
I think the third point would be the same as the first point for most systems, right?
I guess you could keep the r,z inputs you have as overrides, if you really need them, but getting data from the ODS is probably much smoother.

If you want to simplify the calculation to work just in the R-Z plane (which I assume one would do for the initial implementation), then please add a check that phi matches for all the points. It should probably raise `NotImplementedError` if asked to do the calculation for different phi points unless you bother to support that.

omas/omas_physics.py
```diff
@@ -180,6 +181,74 @@ def summary_greenwald(ods, update=True):
...
+
+    """
+     Calculates line-aveage electron dnesity for each time slice and stores them in the summary ods.
+
+    :param ods: input ods
+
+    :param update: operate in place
```
eldond, line 193: Docstring should list all the parameters

omas/omas_physics.py
eldond: `omfit/classes/omfit_omas_d3d.py` already contains `setup_interferometer_hardware_description()`, which loads DIII-D interferometer geometry to the ODS.

eldond (changes requested):

omas/omas_physics.py
eldond: omfit_omas_d3d already has the interferometer geometry.

omas/omas_physics.py
```diff
@@ -202,51 +207,66 @@ def summary_lineaverage_density(ods, R1, R2, Z1, Z2, line_grid=2000, time_index=
...
+    
+    for channel in ods['interferometer']['channel']: 
+    
+        R1 = ods['interferometer']['channel'][channel]['line_of_sight']['first_point']['r']
+        Z1 = ods['interferometer']['channel'][channel]['line_of_sight']['first_point']['z']
+        phi1 = ods['interferometer']['channel'][channel]['line_of_sight']['first_point']['phi']
+        
+        R2 = ods['interferometer']['channel'][channel]['line_of_sight']['second_point']['r']
+        Z2 = ods['interferometer']['channel'][channel]['line_of_sight']['second_point']['z']
+        phi2 = ods['interferometer']['channel'][channel]['line_of_sight']['second_point']['phi']
```
eldond, line 242: You've got to include the third point (it's in the schema). At DIII-D, the third point is the same as the first point, so you can cover it with a simple factor of 2. But in general, the return path can be different (it is like this at KSTAR).

eldond (changes requested):

omas/omas_physics.py
eldond: I realize I mentioned that the first point is the same as the third point for most systems, which I think is probably true. Since that time, I learned more about KSTAR's system, so it's important to support point3 != point1.

jmcclena: @eldond Arg! I used your comment to justify my laziness.

eldond: Yeah, sorry for moving the goalposts.

omas/omas_physics.py
```diff
@@ -180,6 +181,102 @@ def summary_greenwald(ods, update=True):
...
+			
+            zero_crossings = numpy.where(numpy.diff(numpy.sign(numpy.gradient(dist))))[0]
+            i1 = zero_crossings[0]
+            i2 = zero_crossings[-1]
+
+            ne_line = 0.
+            for i in range(i1, i2, numpy.sign(i2-i1)):
+                psival = psi_interp(Zline[i], Rline[i])[0]
+                ne_interp  = scipy.interpolate.splev(psival, tck)
+                ne_line  += ne_interp
```
eldond, line 268: If you used the trapezoid rule, could you get away with fewer points along the line and thus make it faster?
```suggestion
            psival = [psi_interp(Zline[i], Rline[i])[0] for i in range(i1, i2, numpy.sign(i2-i1))]
            ne_interp  = scipy.interpolate.splev(psival, tck)
            ne_line  = numpy.trapz(ne_interp)
```

eldond (approved):
Seems okay now

eldond:
@orso82 it fails with
```
======================================================================
FAIL: test_latest_cocos (test_omas_physics.TestOmasPhysics)
----------------------------------------------------------------------
  File "/home/travis/build/gafusion/omas/omas/tests/test_omas_physics.py", line 452, in test_latest_cocos
    list_structures(imas_version=omas_rcparams['default_imas_version']), threshold=0, write=False, verbose=False
  File "/home/travis/build/gafusion/omas/omas/omas_physics.py", line 1975, in generate_cocos_signals
    assert original == new, 'COCOS signals are not up-to-date! Run `make cocos` to update the omas_cocos.py file.\n' + ''.join(diff)
AssertionError: COCOS signals are not up-to-date! Run `make cocos` to update the omas_cocos.py file.
```
so maybe `make cocos` should be in travis.yaml or something.

orso82:
@eldond the failure indicates that something is wrong with cocos. This should not be fixed with adding 'make cocos' to travis.

eldond:
Maybe the error message needs an update then, because it asked me to run `make cocos` and travis seems to be doing better with that added. But go ahead and revert 205b5e5 if you can do it better.  

Also, why are the tests running on travis-ci.org and travis-ci.com? Can the redundant .org ones be stopped (travis-ci.org is shutting down in favor of .com at the end of the year https://blog.travis-ci.com/2018-05-02-open-source-projects-on-travis-ci-com-with-github-apps).

orso82:
I am quite confident that https://github.com/gafusion/omas/commit/205b5e5cfb4df67c7307c6364232749e662e32ae worked because of the change to omas_cocos.py and not because of the changes to the travis file...

omas/omas_physics.py
```diff
@@ -180,6 +181,102 @@ def summary_greenwald(ods, update=True):
...
+            x1 = R1*numpy.cos(phi1)
+            y1 = R1*numpy.sin(phi1)
+            		
+            R2 = ods['interferometer']['channel'][channel]['line_of_sight'][ifpath[1]]['r']
+            Z2 = ods['interferometer']['channel'][channel]['line_of_sight'][ifpath[1]]['z']
+            phi2 = ods['interferometer']['channel'][channel]['line_of_sight'][ifpath[1]]['phi']
+            x2 = R2*numpy.cos(phi2)
+            y2 = R2*numpy.sin(phi2)
+                
+            xline = (x2-x1) * numpy.linspace(0, 1, line_grid) + x1
```
orso82, line 251:
```suggestion
            xline = numpy.linspace(x1, x2, line_grid)
```

jmcclena: Thats for correcting this. I missed David's comment.

omas/omas_physics.py
```diff
@@ -248,7 +248,7 @@ def summary_lineaverage_density(ods, line_grid=2000, time_index=None, update=Tru
             x2 = R2*numpy.cos(phi2)
             y2 = R2*numpy.sin(phi2)
                 
-            xline = (x2-x1) * numpy.linspace(0, 1, line_grid) + x1
+            xline = numpy.linspace(x1, x2, line_grid)
             yline = (y2-y1) * numpy.linspace(0, 1, line_grid) + y1			
             Rline = (R2-R1) * numpy.linspace(0, 1, line_grid) + R1
             Zline = (Z2-Z1) * numpy.linspace(0, 1, line_grid) + Z1
```
eldond: Okay, but there was more context to this one.

omas/omas_physics.py
```diff
@@ -261,11 +261,9 @@ def summary_lineaverage_density(ods, line_grid=2000, time_index=None, update=Tru
...
             i2 = zero_crossings[-1]
 
-            ne_line = 0.
-            for i in range(i1, i2, numpy.sign(i2-i1)):
-                psival = psi_interp(Zline[i], Rline[i])[0]
-                ne_interp  = scipy.interpolate.splev(psival, tck)
-                ne_line  += ne_interp
+            psival = [psi_interp(Zline[i], Rline[i])[0] for i in range(i1, i2, numpy.sign(i2-i1))]
+            ne_interp  = scipy.interpolate.splev(psival, tck)
+            ne_line  = numpy.trapz(ne_interp)
```
eldond, line 266: @jmcclena does use of trapz allow fewer points to be used along the interf line of sight?

jmcclena: @eldond I'm not quite sure what you mean.

eldond: You were interpolating to get a bunch of points sampling the line and then adding all the values together. Presumably using trapezoid integration gives higher accuracy than boxes for the same number of points, or the same accuracy for fewer points (making it faster). Also, does this work as expected?

jmcclena: Oh, I see. Yes, I'm getting very similar results. Presumably trapz is more accurate. I'm getting slightly faster for the new algorithm.

PR #133, opened 2020-10-17
Avoid in-place multiplication of color list

eldond:
- Was ensuring enough colors for overlays by colors *= nc
- This updates the source color list; repeated calls with the same colors object were resulting in the list of colors elongating until a MemoryError was triggered

### DIFF
--- a/omas/omas_plot.py
+++ b/omas/omas_plot.py
@@ -2694,7 +2694,8 @@ def bolometer_overlay(ods, ax=None, reset_fan_color=True, colors=None, **kw):
     if colors is None:
         colors = [kw.pop('color', None)]
     ci = 0
-    color = (colors * nc)[ci]  # Multiplying list by nc makes sure it's always long enough.
+    colors2 = colors * nc
+    color = colors2[ci]  # Multiplying list by nc makes sure it's always long enough.
     kw.setdefault('alpha', 0.8)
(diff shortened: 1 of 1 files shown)


smithsp:
Maybe https://matplotlib.org/cycler/ would be helpful here?

eldond:
@smithsp it doesn't seem easy to use cycler this time. I want definite colors for each value of a counter, and I may use two loops over the same thing in places.

smithsp:
@eldond Maybe the matplotlib cycler was too much.  I investigated further and it uses `itertools.cycle` underneath, which could be used as:
```python
for i,(l,c) in enumerate(zip(labels,itertools.cycle(colors))):
```
or
```
colors = [c for c in zip(labels, itertools.cycle(color))]
...
   colors[i]
```

eldond:
@smithsp I don't know how to test for how the plots look in terms of using consistent or reasonable colors, so the unit tests won't catch a problem with the wrong coloration due to a mishap with a cycler. So, that increases the time and effort required for redevelopment and testing and I can't do it right now. You're welcome to take over or open an issue if you think it's important. I agree that using a cycler from the start would've been nicer than these lists, and I'll try to remember to do that next time.

smithsp:
@eldond I'm not going to take over.  I just wanted to make sure that you were aware when you have the bandwidth.

PR #142, opened 2021-02-03
Omas plot, physics and added rotation to omas_cocos

TimSlendebroek:


omas/omas_physics.py
```diff
@@ -437,54 +445,48 @@ def summary_total_powers(ods, update=True):
...
+                if 'total_ion_energy' in source_1d:
+                    q_dict['total_heating'] += source_1d['total_ion_energy']
+                    if sources[source]['identifier.index'] in power_dict:
+                        q_dict[index_dict[sources[source]['identifier.index']]] += source_1d['total_ion_energy']
 
     for key, value in power_dict.items():
+        power_dict[key].append(numpy.trapz(q_dict[key], x=vol))
         if numpy.sum(value) > 0:
-            if key == 'total':
+            if key is 'total_heating':
```
orso82, line 481: cannot compare strings with `is`. This will always evaluate to false.

omas/omas_physics.py
```diff
@@ -387,24 +387,31 @@ def summary_taue(ods, update=True):
...
-                * (total_power / 1e6) ** -0.69
+                * (heating_power / 1e6) ** -0.69
                 * r_major ** 1.97
                 * kappa ** 0.78
                 * aspect ** -0.58
                 * isotope_factor ** 0.19
             )  # [s]
-            # print('kap', kappa, 'bt', bt, 'ip', ip, 'ne_vol', ne_vol_avg, 'paux', p_aux, 'aspect', aspect, 'isotope', isotope_factor, 'tau_e', tau_e)
+            if PRINTFLAG:
+                print('kap', kappa, 'bt', bt, 'ip', ip, 'ne_vol', ne_vol_avg, 'paux', p_aux, 'aspect', aspect, 'isotope', isotope_factor, 'tau_e', tau_e)
```
orso82, line 412: ugly. remove PRINTFLAG  and use `printd()`
```suggestion
            for k in ['kappa', 'bt', 'ip', 'ne_vol_avg', 'p_aux', 'aspect', 'isotope_factor', 'tau_e']:
                printd(f'{k}: {eval(k)}', topic='summary_taue')
```

you can then see this output by setting `os.environ['OMAS_DEBUG_TOPIC']='summary_taue'`

orso82 (changes requested):
Looks mostly good. Can you please make sure we run these as part of the regression tests?

omas/omas_physics.py
TimSlendebroek: yeah you are right 
```if 'total_heating' in key```
 is way better

**Comment by TimSlendebroek** [comment]
> 
> 
> Looks mostly good. Can you please make sure we run these as part of the regression tests?
@orso82 
Will do, they were in the regression test but sneakily the sample ods doesnt' have a summary!

File: omas/omas_physics.py [hunk 3]
```diff
@@ -478,7 +478,7 @@ def summary_heating_power(ods, update=True):
     for key, value in power_dict.items():
         power_dict[key].append(numpy.trapz(q_dict[key], x=vol))
         if numpy.sum(value) > 0:
-            if key is 'total_heating':
+            if 'total_heating' in key:
                 ods_n['summary.global_quantities.power_steady.value'] = numpy.array(value)
                 continue
             elif key is 'fusion':
```
orso82: make sure you get all of these: `key is 'fusion'`

TimSlendebroek: I think I caught it in one of my follow up commits

omas/tests/test_omas_physics.py
```diff
@@ -187,6 +187,7 @@ def CPC(ods, kw=jdef, should_RE=False, should_AE=False, warn=False):
 
     def test_sumary_global_quantities(self):
         ods = ODS().sample()
+        ods['summary'] = ODS()
```
orso82, line 190: @TimSlendebroek why is this needed?

omas/omas_physics.py
```diff
@@ -437,54 +439,49 @@ def summary_total_powers(ods, update=True):
...
+                        q_dict[index_dict[sources[source]['identifier.index']]] += source_1d['total_ion_energy']
 
     for key, value in power_dict.items():
+        power_dict[key].append(numpy.trapz(q_dict[key], x=vol))
         if numpy.sum(value) > 0:
-            if key == 'total':
+            if 'total_heating' in key:
                 ods_n['summary.global_quantities.power_steady.value'] = numpy.array(value)
                 continue
+            elif 'fusion' in key:
```
orso82, line 478:
```suggestion
            elif key == 'fusion':
```

omas/omas_physics.py
```diff
@@ -437,54 +439,49 @@ def summary_total_powers(ods, update=True):
...
+                if 'total_ion_energy' in source_1d:
+                    q_dict['total_heating'] += source_1d['total_ion_energy']
+                    if sources[source]['identifier.index'] in index_dict and index_dict[sources[source]['identifier.index']] in q_dict:
+                        q_dict[index_dict[sources[source]['identifier.index']]] += source_1d['total_ion_energy']
 
     for key, value in power_dict.items():
+        power_dict[key].append(numpy.trapz(q_dict[key], x=vol))
         if numpy.sum(value) > 0:
-            if key == 'total':
+            if 'total_heating' in key:
```
orso82, line 475:
```suggestion
            if key == 'total_heating':
```

omas/omas_physics.py
```diff
@@ -336,7 +336,7 @@ def summary_lineaverage_density(ods, line_grid=2000, time_index=None, update=Tru
 
 @add_to__ODS__
 @preprocess_ods('core_profiles', 'core_sources', 'equilibrium')
-def summary_taue(ods, update=True):
+def summary_taue(ods, update=True, PRINTFLAG=False):
```
orso82, line 339:
```suggestion
def summary_taue(ods, update=True):
```

orso82 (commented):
few more details

omas/tests/test_omas_physics.py
TimSlendebroek: Well the sample doesn't have summary, which makes triggering ods.summary_global_quanties() not possible

orso82: Can we generalize this fix and make it so that `summary_global_quantities` adds `ods['summary']` if that's not present?

TimSlendebroek: Yeah that's fair good point!

PR #143, opened 2021-02-10
allow different memory backends in OMAS

orso82:
For much of what we have done so far, IDSs data was stored in OMAS in Python structures, and then moved it in/out of IMAS only when needed. This PR allows OMAS data to be stored directly in IMAS (ie, avoiding copy of data from Python structures and IMAS). Indeed the new storage requires IMAS to be installed.

image

```python
# Define the memory backend when the ODS is created
ods = ODS(backend='imas')

# put in some data
ods['equilibrium.ids_properties.homogeneous_time'] = 1

# use `.ids` to access IMAS data structure
equilibrium_ids_in_imas = ods['equilibrium'].ids

# for example:
in_omas = ods['equilibrium.ids_properties.homogeneous_time']
in_imas = ods.ids.equilibrium.ids_properties.homogeneous_time
```


smithsp:
Mention @olivhoenen .  This is where the IMAS memory version is being developed.

Karel-van-de-Plassche:
Could we leverage the new (just delivered) IMASPy backend paradigm here @orso82? I'm not sure when IO is ready to switch from IMAS Python HLI to IMASPy, but more testers could help this along.

PR #146, opened 2021-03-05
fix zeff calculation with ions

jmcclena:


jmcclena:
@orso82 This bug is causing zeff to be off for all STEP simulations. Could you merge this pull request.

smithsp (approved):

Issue #149, opened 2021-03-26
Ci testing examples also


dboeckenhoff:
Hi, when diving into omas and running examples some do not work, e.g.

`$ python omas/examples/omas_dynamic_machine.py `

> Traceback (most recent call last):
>   File "omas/examples/omas_dynamic_machine.py", line 12, in <module>
>     from omfit_classes.omfit_eqdsk import OMFITgeqdsk
> ModuleNotFoundError: No module named 'omfit_classes'

This requires omfit which is not part of the omas dependencies. I propose to include the execution of examples in a ci stage.

smithsp:
I do not think that `omas` will ever depend on `omfit_classes`, as that would introduce a circular dependence, where `omas` is a more fundamental package and OMFIT is more of an accumulator package.  So the example, while indicative of how OMFIT and omas could interact, should be robust to `omfit_classes` not being installed, and the CI should check the examples.  Would you be willing to make the example robust to `omfit_classes` not being installed (it's really only needed for that last part of the example) and create a PR?

orso82:
@dboeckenhoff the examples are already part of the CI (see `omas/omas/tests/test_omas_examples.py`). However, the examples that depend on the `omfit_classes` are skipped if these are not installed. I have not yet added `omfit_classes` to the list of requirements simply because we still need to get around to make a conda package for them (they are available on PYPI). For the time being, you  can install the omfit-classes with `pip install --upgrade omfit_classes`

@smithsp I think the circular dependency should not be a problem, since that's something that `pip` or `conda` should figure out.

dboeckenhoff:
@orso82 sorry, yes I did not see test_omas_examples.py - shame on me. By the way: do you guys know (pytest)[https://docs.pytest.org/en/stable/]? It has enormous powers on top of unittest and also integrates e.g. doctests easily.

@orso82 @smithsp I hope with all these issues I am not giving the wrong impression. I really appreciate your work here. I am happy about this intense support from you guys! Thanks again.

orso82:
@dboeckenhoff I have given you access to the repo (master branch is protected)
We appreciate your feedback and we're always happy to learn and improve. I am not familiar with pytest :)

PR #152, opened 2021-03-29
Added note to omas_dynamic_machine example for required imports

dboeckenhoff:
Solving #149

### DIFF
--- a/omas/examples/omas_dynamic_machine.py
+++ b/omas/examples/omas_dynamic_machine.py
@@ -8,11 +8,27 @@ Only the data that is queried in the ODS will be loaded.
 """
 
 import os
+import logging
 from omas import *
-from omfit_classes.omfit_eqdsk import OMFITgeqdsk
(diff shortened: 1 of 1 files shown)


dboeckenhoff:
Ci failed in build stage with unit tests. Locally on my machine tests run fine and pass. Not sure what is the problem. Any idea?

orso82:
@dboeckenhoff I know that your commit was solicited by @smithsp comment, but I have decided to revert your commit and simply added a note at the beginning of the example that states that to run the MDSplus and omfit_classes are required. I'd really like to keep the examples as focused as possible on the OMAS library itself.

orso82:
@dboeckenhoff I don't know why the regression tests are failing now... I am looking into it

orso82:
@dboeckenhoff it has to be something in the environment, since an empty commit to a separate branch that was previously ok, also shows a fail:
image

PR #164, opened 2021-06-01
Efficiently handle uncertain data with new ouarray class

orso82:


omas/omas_utils.py
```diff
@@ -7,6 +7,213 @@
...
+    def std_dev(self):
+        '''
+        Return standard deviation for symmetric uncertainty bounds
+        NOTE: This method will raise an error if uncertainty is not symmetrical
+        '''
+        if self._error_upper is None:
+            return self.nominal_value * 0.0
+        elif self._error_lower is None:
+            return self.error_upper
+        else:
```
smithsp, line 139: What about the case where `self.error_upper==self.error_lower`?

omas/omas_utils.py
```diff
@@ -7,6 +7,213 @@
...
+
+
+def nominal_values(obj):
+    if isinstance(obj, ouarray):
+        return obj.nominal_value
+    else:
+        return uncertainties.unumpy._orig_nominal_values(obj)
+
+
+uncertainties.unumpy.nominal_values = nominal_values
```
smithsp, line 31: Does this monkey-patching imply a certain order of importing `omas` and `uncertainties`?

smithsp (approved):

omas/omas_utils.py
orso82: Good point, but after some thought I think it is better to raise an error, the point being that if the error is symmetrical then only `error_upper` should be filled. It's better to raise these issues always, rather than having things pass in some cases but not others.

smithsp: As long as we thought about it, and made an active decision, I am OK.

omas/omas_utils.py
orso82: This works like all other monkey-patches. In OMFIT for example, we should import omas early on, so after that, the monkey patched version of `nominal_values` and `std_devs` will be imported everywhere else.

smithsp: Sure, OMFIT will do it right, but do you want to check for `uncertainties` in `sys.modules` here in `omas`, and warn or error if already imported?

PR #166, opened 2021-07-28
D3d cer fetch

kfoss546:
Added etendue values to bolometer hardware


orso82:
@kfoss546 @smithsp I have simply renamed the d3d_cer.py file. This change will also need to be propagated in the OMFITprofiles fetching routines

PR #169, opened 2021-08-31
Added d3d machine mappings for ECH

AreWeDreaming:
Beam spot size and curvature radius still WIP.
Added test and test data.
Will convert out of draft mode once I hear beak from ECH group about beam spot size and curvature,

### DIFF
--- a/omas/machine_mappings/d3d.py
+++ b/omas/machine_mappings/d3d.py
@@ -327,6 +327,92 @@ def pf_active_coil_current_data(ods, pulse):
             data_norm=1.0,
         )
 
+#================================   
+@machine_mapping_function(__regression_arguments__, pulse=170325)
+def ec_launcher_active_hardware(ods, pulse):
+    setup = '.ECH.'
+    # We need three queries in order to retrieve only the fields we need
+    # First the amount of systems in use
+    query = {'NUM_SYSTEMS' : setup + 'NUM_SYSTEMS'}
+    num_systems = mdsvalue('d3d', treename='RF', pulse=pulse, TDI=query).raw()['NUM_SYSTEMS']
+    query = {}
+    # Second query the used systems to resolve the gyrotron names
+    for system_no in range(1, num_systems + 1):
+        cur_system = f'SYSTEM_{system_no}.'
+        query[f'GYROTRON_{system_no}'] = setup + cur_system + 'GYROTRON.NAME'
+        query[f'FREQUENCY_{system_no}'] = setup + cur_system + 'GYROTRON.FREQUENCY'
+        for field in ['LAUNCH_R', 'LAUNCH_Z', 'PORT']:
+            query[field + f'_{system_no}'] = setup + cur_system + f'antenna.{field}'
+    systems = mdsvalue('d3d', treename='RF', pulse=pulse, TDI=query).raw()
+    query = {}
+    gyrotron_names = []
+    for system_no in range(1, num_systems + 1):
+        if(len(systems[f'GYROTRON_{system_no}']) == 0):
+            """
+            If nothing is connected to this system the gyrotron name is blank.
+            """
+            continue
+        gyrotron = systems[f'GYROTRON_{system_no}']
+        gyrotron_names.append(gyrotron)
+        gyr = gyrotron.upper()
+        gyr = gyr[:3]
+        for field in ['STAT', 'XMFRAC', 'FPWRC', 'AZIANG', 'POLANG']:
+            query[field + f'_{system_no}'] = setup + f'{gyrotron.upper()}.EC{gyr}{field}'
+            if(field in ['XMFRAC', 'FPWRC', 'AZIANG', 'POLANG']):
+                query["TIME_" + field + f'_{system_no}'] = ("dim_of("
+                        + query[field + f'_{system_no}'] + "+01)")
+    gyortrons = mdsvalue('d3d', treename='RF', pulse=pulse, TDI=query).raw()
+    for system_no in range(1, num_systems + 1):
+        system_index = system_no - 1
+        if gyortrons[f'STAT_{system_no}'] == 0:
+            continue
+        launchers = ods['ec_launchers.launcher']
+        launchers[system_index]['identifier'] = launchers[system_index]['name'] = gyrotron_names[system_index]
+        launchers[system_index]['launching_position.r'] = np.atleast_1d(systems[f'LAUNCH_R_{system_no}'])
+        launchers[system_index]['launching_position.z'] = np.atleast_1d(systems[f'LAUNCH_Z_{system_no}'])
+        launchers[system_index]['launching_position.time'] = np.zeros(
+            launchers[system_index]['launching_position.r'].shape)
+        phi = np.deg2rad(float(systems[f'PORT_{system_no}'].split(' ')[0]))
+        phi = -phi - np.pi / 2.0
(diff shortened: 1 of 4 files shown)


omas/tests/machine_mappings/test_d3d.py
```diff
@@ -0,0 +1,38 @@
...
+
+class TestD3D(unittest.TestCase):
+
+    def test_d3d_ECH_machine_mapping(self):
+        '''
+        Test the machine mapping of the DIII-D ECH against the a .json file created with fetchGyrotron from the OMFIT TORAY module
+        '''
+        pristine_ods = omas.load_omas_json(os.path.join('tests', 'test_data', 'd3d_ECRH_machine_mapping.json'))
+        test_ods = omas.ODS()
+        tunnel_mds('d3d', 'RF')
```
orso82, line 16: not sure why this is necessary

omas/tests/machine_mappings/test_data/d3d_ECRH_machine_mapping.json
```diff
@@ -0,0 +1,312 @@
+{
```
orso82, line 1: @AreWeDreaming we have been placing these files under `omas/samples`
Also, you could update `omas/omas_sample.py` to support `ods.sample("ec_launchers")`

omas/tests/machine_mappings/test_d3d.py
```diff
@@ -0,0 +1,38 @@
...
+        ec_launcher_active_hardware(test_ods, 170325)
+        # test_ods.open('d3d', 170325)
+        fields = ['launching_position.r', 'launching_position.z', 'launching_position.phi', 
+                  'frequency.data', 'power_launched.data', 'mode.data', 'steering_angle_tor.data',
+                  'steering_angle_pol.data']
+        prefix = 'ec_launchers.launcher'
+        for launch in pristine_ods[prefix]:
```
orso82, line 23: You can compare two ods with:
```python
ods1.diff(ods2)
```

omas/machine_mappings/d3d.py
```diff
@@ -327,6 +327,92 @@ def pf_active_coil_current_data(ods, pulse):
             data_norm=1.0,
         )
 
+#================================   
+@machine_mapping_function(__regression_arguments__, pulse=170325)
+def ec_launcher_active_hardware(ods, pulse):
+    setup = '.ECH.'
+    # We need three queries in order to retrieve only the fields we need
```
orso82, line 334: the comment about the third query is missing :)
```
# First the amount of systems in use
...
# Second query the used systems to resolve the gyrotron names
...
# Third ???
```

omas/tests/machine_mappings/test_d3d.py
AreWeDreaming: I don't think this is going to work here. The TORAY generated ods only has one timepoint while the machine mapping routine loads all timepoints. Also, they don't hold the same fields. Some time fields are missing in the TORAY sample file, the time units are in milliseconds and there are some code parameters in the TORAY file that are not loaded with the machine mapping. For the sample, it would be most likely better to only have one timepoint to save git space.

omas/machine_mappings/d3d.py
AreWeDreaming: I was hoping to get away with 2 queries. Must have forgotten to add that comment when I resigned to use three calls xD

omas/tests/machine_mappings/test_d3d.py
orso82: @AreWeDreaming time must always be in seconds in an ODS

AreWeDreaming: Yes, but this is a bug in the TORAY module. I didn't want to touch that at all so I could use its output for my test. Arguably now that everything has been validated once, it might be better to create an ODS with the new machine mapping routines and use that for future verification. Should I do that?

omas/tests/machine_mappings/test_d3d.py
AreWeDreaming: The tunnel was necessary to call the machine mapping function directly. If I used the ods.open() statement my debugger cannot find d3d.py. I will remove it now and use ods.open instead.

omas/tests/machine_mappings/test_data/d3d_ECRH_machine_mapping.json
AreWeDreaming: Do you mean edit the `omas_core.py` function sample?
```
sample(self, ntimes=1, homogeneous_time=None)
```
To something like:
```
sample(self, pr="All", ntimes=1, homogeneous_time=None)
```
I have added the function `ec_launchers` to `omas_sample.py` and then run `omas.sample(1)` to fetch all sample data including the `ec_launchers`.

omas/tests/machine_mappings/test_d3d.py
AreWeDreaming: Removing it produces the following error:
This produces the following, unrelated error
```
  File "/mnt/c/Users/Severin/git/omas/omas/tests/machine_mappings/test_d3d.py", line 17, in test_d3d_ECH_machine_mapping
    test_ods.load('d3d', 170325)
  File "/mnt/c/Users/Severin/git/omas/omas/omas_core.py", line 2081, in load
    results = eval('load_omas_' + ext)(*args, **kw)
  File "/mnt/c/Users/Severin/git/omas/omas/omas_machine.py", line 950, in load_omas_machine
    machine_to_omas(ods, machine, pulse, location, options, branch)
  File "/mnt/c/Users/Severin/git/omas/omas/omas_machine.py", line 207, in machine_to_omas
    f"Experiment data {data.shape} does not fit in `{location}` [{', '.join([':'] * location.count(':') + mapped.get('COORDINATES', []))}]"
Exception: Experiment data (1, 1, 331) does not fit in `equilibrium.time_slice.:.profiles_2d.:.grid_type.index` [:, :]
```
This is very weird because I can run ods.load('d3d', 170325) in other scripts. I don't know how to fix this. I submitted a commit of my current head for debugging purposes.

AreWeDreaming: Ok, I was using `omas.load()` instead of `omas.open()`. So this is fixed, but now I get this interaction:
If I do:
```
pristine_ods = omas.ODS()
pr = load_omas_json(os.path.join('omas','samples', 'd3d_ECRH_machine_mapping.json'))['ec_launchers']
pristine_ods['ec_launchers'].update(pr)
```
or
```
pristine_ods = load_omas_json(os.path.join('omas','samples', 'd3d_ECRH_machine_mapping.json'))
```
I can use `pristine_ods['ec_launchers.launcher[0]'],time()` to get the time base.
However when I do:
```
pristine_ods = omas.ODS()
pristine_ods = pristine_ods.sample()
```
with
```
@add_to_ODS
def ec_launchers(ods):
    from omas import load_omas_json

    pr = load_omas_json(os.path.join('omas','samples', 'd3d_ECRH_machine_mapping.json'))['ec_launchers']
    ods['ec_launchers'].update(pr)
    return ods
```
I can use `pristine_ods['ec_launchers.launcher[0]'],time()` returns `None`.

orso82:
I am very glad that you got to talk to Neal I see regression tests failing, which is not surprising since the GH runners do not have MDS+ installed nor access to DIII-D. I suggest you move the test that you have created to an example, or simply remove it all-together.

PR #172, opened 2021-10-15
2d mapping for equilibrium

AreWeDreaming:
- Adds mapping routines for 2d_profiles quantities in the equilibrium.
- Automatically derives magnetic field components if missing from ODS.
- New ScatterInterpolator class to provide a uniform interface to various scipy interpolator


AreWeDreaming:
Work in progress and not ready to merge. Requires testing and mapping between magnetic coordinates needs to be added.

AreWeDreaming:
The derivation of the magnetic field was tested for the ECE line of sight. Differences w.r.t. the magnetic field from omfit.geqdsk are smaller than 1%. The vacuum B_t field is not time-resolved in the current version of the equilibrium machine mappings. As a replacement profiles1d.f[-1] is used to calculate B_t outside of closed field lines.

PR #176, opened 2021-11-05
Fixed argument order in ece d3d mapping

AreWeDreaming:
The order of the arguments for electron_cyclotron_emission_data was wrong making it unable to fetch data. This commit fixes this.


AreWeDreaming:
Also fixed time which was incorrectly in ms instead s.

PR #179, opened 2021-11-22
Include fusion power in neutrons

bclyons12:
Strangely, `summary.fusion.power` is just the power coupled to the plasma (according the IMAS schema). This updates it, assuming it's 100% D-T fusion.

I'm not sure if we should worry about a contribution from D-D fusion. I think there's an O(1%) correction from that?

### DIFF
--- a/omas/omas_physics.py
+++ b/omas/omas_physics.py
@@ -879,6 +879,7 @@ def summary_heating_power(ods, update=True):
                 continue
             elif key == 'fusion':
                 ods_n['summary.fusion.power.value'] = numpy.array(value)
+                ods_n['summary.fusion.neutron_power_total.value'] = (14.1 / 3.5) * numpy.array(value)
                 continue
             ods_n[f'summary.heating_current_drive.{key}[0].power.value'] = numpy.array(value)
 


TimSlendebroek:
@bclyons12 

Thermal TT and DD might not be super significant but beam-target is a very significant fraction (for example the DT JET campaigns).

What I find weird is that they do split up all the neutron_fluxes but not the neutron powers:
+ summary.fusion.neutron_fluxes.tt.beam_thermal
+ summary.fusion.neutron_fluxes.tt.thermal.source

While there is only (for the powers):
+ summary.fusion.neutron_power_total


Maybe something to bring up to the IMAS dataschema people @bclyons12  @orso82 ?

If that is added we could do summary.fusion.neutron_power_total = the sum of all the individual components and fill in DT in TGRYO_step

bclyons12:
@TimSlendebroek Maybe because each fusion reaction results in monoenergetic neutrons, so if you have the neutron flux, you can compute the power?

TimSlendebroek:
@bclyons12 
I agree but that is true for almost every derived quantity (that is almost the definition of a derived quantity) and there are many of those in the IMAS dataschema. 

Practically on a now time-frame what do you think about adding a neutron_power_sum() function that sums up all contributions and put the dt neutron flux in it's specific location. Something tells me that this isn't super obvious for beam target though (conservation of energy?).

PR #180, opened 2021-11-22
Fix call to summing j_actuators

bclyons12:
Per change made in 6f599eaebc8c1bcf0221697d0dc1ca0c6d41aeba


bclyons12:
Bug fix needed by @daveweisberg

PR #182, opened 2021-12-06
Fixed time units and data validity time base for DIII-D Interferometry

AreWeDreaming:
Time of data now in seconds instead of milliseconds. Data validity has a different time base than the data in MDS+. For the IDS the time base needs to be consistent. This is now solved using nearest-neighbor interpolation.


orso82:
I am going to close this PR. The commits relevant to the interferometer fix have been cherry-picked into master.
The other commits are already part of PR #172

PR #185, opened 2022-01-06
map EFIT AAxGAM to IMAS geometric_coefficients

orso82:
to be tested with the OMFIT branch mse_IMAS

### DIFF
--- a/omas/machine_mappings/nstxu.py
+++ b/omas/machine_mappings/nstxu.py
@@ -378,6 +378,29 @@ def mse_data(ods, pulse, MSE_revision="ANALYSIS", MSE_Er_correction=True):
     making
         tan(gamma_cor) ~ (big positive) - (small negative)
     Thus, tan(gamma_cor) is more positive than the reported tan(gamma)
+
+
+    mapping between IMAS geometric_coefficients and EFIT AAxGAM
+    coeffs0: AA1
+    coeffs1: AA8
+    coeffs2: AA2
+    coeffs3: AA5
+    coeffs4: AA4
+    coeffs5: AA3
+    coeffs6: 0
+    coeffs7: AA7
+    coeffs8: AA6
+
+    mapping between EFIT AAxGAM and IMAS geometric_coefficients
+    AA1: coeffs0
+    AA2: coeffs2
+    AA3: coeffs5
+    AA4: coeffs4
+    AA5: coeffs3
+    AA6: coeffs8
+    AA7: coeffs7
+    AA8: coeffs1
+    AA9: does not exist
     """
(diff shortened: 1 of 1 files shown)


orso82:
NOTE by @jmcclena : I am unsure whether the units of D1,D2,D3 are the same between IMAS and EFIT, and the signs of all the coefficients should depend on cocos.

jmcclena:
From what I can tell the units and cocos should be: 

0: AA1 - None, [1/T]
1: AA8 - None, [1/T]
2: AA2 - TOR, [1/T]
3: AA5 - None, [m/V]
4: AA4 - None, [1/T]
5: AA3 - None, [1/T]
6: 0 - TOR, [1/T]
7: AA7 - None, [m/V]
8: AA6 - None, [m/V]

I'm not sure how to set the cocos for a mixed array. Could you do something like this:

`_cocos_signals['mse.channel:.active_spatial_resolution:.geometric_coefficients.1']=  None`

orso82:
It's messy that the array has mixed units and COCOS. We could think of supporting mixed arrays by defining an array of cocos transform for that quantity. However, for now I would suggest that we keep things simple and just explicitly handle the cocos conversion for each of the array elements.

@jmcclena one more piece of information that we need is to know what COCOS does EFIT use. In other words, how is AA2 defined in EFIT? Is that with respect to the toroidal angle or with respect to the current or with respect to the magnetic field?

jmcclena:
Okay, I looked into this more. The MSE AA2 angle coefficient in EFIT is a cocos that has a counter-clockwise toroidal angle from the topview  (e.g. cocos=1).  This coefficient is: sin(alpha), where alpha is the angle from beamline to the toroidal angle. As such, AA2 ends up being positive for DIII-D MSE values. https://diii-d.gat.com/diii-d/Mse69config

orso82:
Thank you for taking the time to look into this @jmcclena . I think now we have all of the information that we need.

jmcclena:
@orso82 Are we waiting for anything else to merge this into master?

orso82:
One more thing to check is that the NSTX(-U) MSE mapping use COCOS 11 (it may still be ok if they use COCOS 1).
Then we need to coordinate this merge with the corresponding one in OMFIT https://github.com/gafusion/OMFIT-source/pull/5787 also in this case, if EFIT uses COCOS 1 and IMAS uses COCOS 11, then we may be ok in terms of COCOS. Still the coordination between the two merges is necessary since we shuffled the mapping between IMAS geometric_coefficients and EFIT AAxGAM. One option to make this fail proof would be to update the OMFIT PR and use different mappings depending on the version of OMAS.

PR #186, opened 2022-01-10
Windows compatibility via removal of `pwd` module

Reksoatr:
OMAS can be successfully built and installed on Windows machines (for example, through the Anaconda package manager) save for the dependency of OMAS on the `pwd` module, which is only available on UNIX machines. I propose to fork OMAS into a 'Windows compatible' branch where this dependency on `pwd` (which only shows up once in omas_setup.py and once in omas_utils.py) is removed

### DIFF
--- a/omas/omas_setup.py
+++ b/omas/omas_setup.py
@@ -27,7 +27,7 @@ OMAS v%s only runs with Python 3.6+ and you are running Python %s
         % (__version__, '.'.join(map(str, sys.version_info[:2])))
     )
 
-import pwd
+#import pwd
 import glob
 import json
 import copy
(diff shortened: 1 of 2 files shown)


orso82:
@Reksoatr we certainly do not want to maintain two versions of the source code, but you are welcome to test (bot in `omas_setup.py` and once in `omas_utils.py`) with something like
```python
if platform.system() == 'Windows':
   ...
else
   import pwd
end
```

orso82:
`pwd` is not used extensively and It would be ideal if you could substitute the Linux/MAC `pwd` dependency and coding with something that work under Windows 👍

Reksoatr:
I think I've figured out a good fix to bypass the `pwd` module check if the OS turns out to be Windows. I'll open a new Pull Request with those changes shortly.

PR #188, opened 2022-02-14
Extend d3d machine mapping data

eldond:


eldond:
@orso82 I put in COCOS=1 (copying IP COCOS) in d3d.py, but it wrote COCOS 11 in the json. Is this right? Because I'm telling it I'm inputting in COCOS 1 and it will interpret into COCOS 11?

orso82:
@eldond the data that ends up in the ODS has to be in COCOS 11. Technically you could do that conversion either in the Python or in the JSON. If you are using Python, I suggest that you do that conversion in Python, and then simply report COCOS 11 in the JSON, rather than the opposite.

PR #189, opened 2022-03-15
add mse cocos

jmcclena:


orso82:
@jmcclena have you tested to see if this works?

jmcclena:
hmm... I was hopeful it would as it didn't crash omas. I just tested it though and it does not work :(.

jmcclena:
I tried adding 

```
 944                     elif isinstance(transform,numpy.array):
 945                         norm = np.ones(len(transform))
 946                         for itf,tf in enumerate(transform):
 947                             norm[itf] = omas_physics.cocos_transform(self.cocosio, self.cocos)[tf]
```

to `omas_core.py`, but that didn't seem to fix it

jmcclena:
@orso82 Okay, this pull request works for me 

```
from omas import *
import numpy as np
ods = load_omas_pkl('ods.pkl')

ods['mse.channel.0.active_spatial_resolution.0.geometric_coefficients'] = np.ones(9)
with omas_environment(ods,cocosio=2):
    print(ods['mse.channel.0.active_spatial_resolution.0.geometric_coefficients'])
```
returns 
```
[ 1.  1. -1.  1.  1.  1. -1.  1.  1.]

```

orso82:
Really nice job @jmcclena ! 🥇 
As a refinement, could we consider using a list instead of numpy array to define the transform?

orso82:
Great job @jmcclena !

PR #192, opened 2022-04-18
git issues removed last tested version...

TimSlendebroek:


TimSlendebroek:
@orso82 can you push this one , I screwed up with the last one due to having to reset my branch multiple times before pushing after testing my fix 😞

PR #193, opened 2022-05-06
Jmcclena comp

jmcclena:
This pull request adds a few pointnames to d3d machine_mapping and gives first implementation of incorporating compensation corrections to signals. 

@orso, I'd be grateful any recommendations for improving the compensation strategy.


omas/machine_mappings/d3d.py
```diff
@@ -1022,6 +1061,23 @@ def magnetics_probes_data(ods, pulse):
...
 
+    
+    for compfile in ['btcomp','ccomp','icomp']
+        comp = get_support_file(OMFITd3dcompfile, support_filenames('d3d', compfile, pulse))
+        for shot in comp:
+            if pulse > compshot:
+                compshot = shot
+        for compsig in comp[compshot]:
+            compsig_data = ptdata2(f"{compsig}",{pulse})
+            compsig_time = dim_of(ptdata2(f"{compsig}",{pulse}),0)/1000
```
orso82, line 1072: @jmcclena I don't see a `ptdata2` function as part of this PR

smithsp: `ptdata2` is a server side tdi function.

jmcclena: Thanks. It was untested and poor guessing on my part.  Fixed in [26f2220](https://github.com/gafusion/omas/pull/193/commits/26f2220ff0f541b673d20a4e75105116aba48723)

orso82 (approved):
This looks good! Lots of great work here. It's not easy to make everything work out.
You can run `make format` at the OMAS repository root to format all files.
Consider removing the kEQDSK file that is not used and probably should not be in OMAS anyways. I do not have much comments on how the compensation is done. Looks good to me. Is there something that was bothering you specifically? Maybe slow?

omas/machine_mappings/support_files/d3d/keqdsk_EFIT01
```diff
@@ -0,0 +1,158 @@
+ &IN1
```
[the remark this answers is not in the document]
jmcclena: Okay, I can move it to one of the those places.

jmcclena:
Its not terribly slow. Approximately 8s to generate the constraints plus 0.25s per kfile generated (which I assume a lot of that is in duplicating the file). If there is any obvious way it could be speed up, let me know. 

I haven't found a good place for the loop voltage any ideas? Maybe its considered a passive loop or flux loop?

orso82:
If you parse the kEQDKS once, then you should be able to deepcopy it (not duplicating) and that should be faster (since you'll skip the parsing step).

orso82:
Something like
```python
template_kEQDSK.load() # force a load
for k in range(10):
    kEQDSK=copy.deepcopy(template_kEQDSK)
```

smithsp:
@orso82 
There was also a question about loop voltage.  Do you recall where you put that for NSTX?

orso82:
I see a bunch of leftover print statements. @jmcclena please let me know when this is ready for final review and merge.

jmcclena:
@orso82 This branch is ready for a final review when you have time. Originally, I was hoping to get vloop into omas equilibrium constraints before merging. However, I suspect the correct location for the loop voltage is pf_active since I found written in pf_passive: "any connected loops are expressed as active coil circuits with no power supply attached".  I think this will require some special treatment, as vloop information isn't saved in the mhdin.dat file, and would be better as a seperate pull request.

jmcclena:
Actually, now that I'm thinking about this again. I need to correct the DIII-D error calculation for diagnostics with integrator errors before merging.

jmcclena:
will do

jmcclena:
@orso82, apparently my concerns with integration errors were unwarranted in this case. I did fix the compensation calculation and it is ready for a final review.

PR #197, opened 2022-06-06
test branch for Joey to work on k-files generation for NSTX

jmcclena:
@orso82 This is a branch for testing a fix for fixing issue https://github.com/gafusion/OMFIT-source/issues/6061 by adding in dummy values if mds+ can't be connected to. As validity doesn't exist for pf_active signals, what do you think is the best way to implement some pass if certain coils can't be retrieved from mds+.

### DIFF
--- a/omas/machine_mappings/_common.py
+++ b/omas/machine_mappings/_common.py
@@ -211,12 +211,13 @@ def fetch_assign(
     """
     t = None
     TDIs = []
-
+    
     if isinstance(channels, str):
         channels = ods1[channels]
 
     for stage in ['fetch', 'assign']:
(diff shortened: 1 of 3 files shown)


jmcclena:
Hmmm... if looks like the signals do not exist below discharge:133139

orso82:
I don't think the issue is with the data fetching code itself. The question is: why are we looking for 18 signals when there are only 15 available? are we using the right signals.dat/mhdin.dat files?

smithsp:
@jmcclena doesn't like this approach.

smithsp:
Superceded by work in #193

PR #202, opened 2022-06-24
Fix loading of data with `verbose=False`

### LINKED ISSUE #201: Fail to load IMAS data with `verbose=True`
I'm trying to use `omas` to work with IMAS data, but it somehow fails to load the data when I set `verbose=True` in `omas.load_omas_imas`. In the snippet below you can see that with the verbose flag on, it loads the expected profiles. But with verbose flag off, the core profiles etc are not loaded.

I think I found the issue and will submit a PR to fix it.

```python
>>> from omas import load_omas_imas

>>> ods1 = load_omas_imas(
>>>     machine='jet',
>>>     run=8000,
>>>     pulse=94875,
>>>     user='g2ssmee',
>>>     consistency_check=False,
>>>     verbose=True,
>>> )
Opening existing IMAS data for user='g2ssmee' machine='jet' pulse=94875 run=8000
dataset_description.imas_version is missing: assuming IMAS version 3.34.0
got IDS
- amns_data                  IDS is empty
- barometry                  IDS is empty
...
- waves                      IDS is empty
- workflow                   IDS is empty
summary.volume_average.zeff.value: 100%|█| 73287/73287 [01:28<00:00, 831.81it/s]

>>> print(ods1.keys())
['core_profiles', 'core_sources', 'core_transport', 'dataset_description', 'equilibrium', 'summary']

>>> ods2 = load_omas_imas(
>>>     machine='jet',
>>>     run=8000,
>>>     pulse=94875,
>>>     user='g2ssmee',
>>>     consistency_check=False,
>>>     verbose=False,
>>> )

>>> print(ods2.keys())
['dataset_description']
```

stefsmeets:
This PR fixes a bug where the paths are only loaded with `verbose=True`.

Closes #201

### DIFF
--- a/omas/omas_imas.py
+++ b/omas/omas_imas.py
@@ -452,7 +452,7 @@ def infer_fetch_paths(ids, occurrence, paths, time, imas_version, verbose=True):
                     print(f'* {ds.ljust(ndss)} IDS has data ({len(getattr(ids, ds).time)} times)')
                 except Exception as _excp:
                     print(f'* {ds.ljust(ndss)} IDS')
-                fetch_paths += filled_paths_in_ids(ids, load_structure(ds, imas_version=imas_version)[1], [], [], requested_paths)
+            fetch_paths += filled_paths_in_ids(ids, load_structure(ds, imas_version=imas_version)[1], [], [], requested_paths)
 
         else:
             if verbose:


orso82:
Thank you for the fix!

PR #207, opened 2022-08-05
BUGFIX: Plot wall...vessel as anular and elements 

kripnerl:
- add plotting for vessel annular and vessel elements
- annular tested on actual data
- elements not tested
- https://gafusion.github.io/omas/schema/schema_wall.html


orso82:
Looks good! Thank you for contributing!

PR #209, opened 2022-08-17
Add option to remove integrator drift

odstrcilt:
Flux loop signals and magnetics probe signals can have nonnegligible errors caused by the analog integrator drift. Assuming that the drift is linear in time, this will set signals back to zero after the end of the discharge.

### DIFF
--- a/omas/omas_physics.py
+++ b/omas/omas_physics.py
@@ -212,10 +212,9 @@ def equilibrium_ggd_to_rectangular(ods, time_index=None, resolution=None, method
         profiles_2d['grid.dim2'] = z
     return ods_n
 
-
 @add_to__ODS__
 def equilibrium_form_constraints(
-    ods, times=None, default_average=0.02, constraints=None, averages=None, cutoff_hz=None, update=True, **nuconv_kw
+    ods, times=None, default_average=0.02, constraints=None, averages=None, cutoff_hz=None,rm_integr_drift_after=None, update=True, **nuconv_kw
 ):
     """
     generate equilibrium constraints from experimental data in ODS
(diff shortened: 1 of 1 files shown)


orso82:
@odstrcilt perhaps you forgot to commit the `remove_integrator_drift` function?

odstrcilt:
@orso82 do you have any guess what is causing the regression test to fail? It does not seem to be related to any changes done by me.

smithsp:
@odstrcilt It looks like you ended up with a branch of omas that depends on a new DD from IMAS, where the EC launchers have changed structure.  @orso82 Does this imply that we need to address the launchers on omas now?

orso82:
Yes @smithsp we made changes on a OMFIT branch, but we still need to address this on the OMAS side. Mostly it's about the regression tests and machine mappings.

smithsp:
Need to merge with `master`.

smithsp:
Still need to merge with master branch.

smithsp:
@odstrcilt 
When @orso82 originally looked into fetching NSTX data using omas, he had gone to great lengths to compare what was fetched to the existing k-files.  How do these changes affect that comparison and validation?

orso82:
@odstrcilt in case it's needed, you can look at the https://github.com/gafusion/OMFIT-source/blob/unstable/modules/EFITtime/SCRIPTS/NSTX/update_keqdsk_from_omas.py script, and in particular use the `OMFITkeqdsk.compare_omas_constraints` function.

smithsp:
Ping @odstrcilt to compare to previous k-files.

odstrcilt:
I have tested shot  133964 at 285ms

Here is compared OMAS case with and without drift substraction
image

Here is OMAS with drift subtraction compared to EFIT01 kfile
image

Here is OMAS without drift subtraction compared to EFIT01 kfile
image

I will try to plot also timetraces...

odstrcilt:
@orso82 can we merge it?

orso82:
A

PR #214, opened 2022-10-06
Fixed DIIID CO2 geo with OMFITprofiles

AreWeDreaming:
The DIII-D mapping used arbitrary endpoints for the CO2 interferometer. The machine mapping was updated with the coordinates from OMFITprofiles.

### DIFF
--- a/omas/machine_mappings/d3d.py
+++ b/omas/machine_mappings/d3d.py
@@ -508,7 +508,8 @@ def interferometer_hardware(ods, pulse):
     """
     Loads DIII-D CO2 interferometer chord locations
 
-    The chord endpoints ARE NOT RIGHT. Only the R for vertical lines or Z for horizontal lines is right.
+    The chord endpoints are approximative. They do not take into account 
+    the vessel wall contour of the shot. The values have been taken from OMFITprofiles.
 
     Data sources:
     DIII-D webpage: https://diii-d.gat.com/diii-d/Mci accessed 2018 June 07 by D. Eldon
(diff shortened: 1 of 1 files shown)


omas/machine_mappings/d3d.py
```diff
@@ -523,15 +523,16 @@ def interferometer_hardware(ods, pulse):
     ods['interferometer.channel.0.identifier'] = ods['interferometer.channel.0.name'] = 'r0'
     r0 = ods['interferometer.channel.0.line_of_sight']
     r0['first_point.phi'] = r0['second_point.phi'] = 225 * (-np.pi / 180.0)
-    r0['first_point.r'], r0['second_point.r'] = 3.0, 0.8  # These are not the real endpoints
```
orso82, line 526: @AreWeDreaming you may want to remove the comment above: "The chord endpoints ARE NOT RIGHT. Only the R for vertical lines or Z for horizontal lines is right.".

AreWeDreaming:
I will. I also just noticed that the CO2 is double pass. Per IMAS convention it should [have 3 points](https://gafusion.github.io/omas/schema/schema_interferometer.html).

orso82:
About the DIII-D interferometer being double pass, is then the density reported in MDS+ already halved? What's the convention in IMAS for that?

AreWeDreaming:
"Line integrated density, possibly obtained by a combination of multiple interferometry wavelengths. Corresponds to the density integrated along the full line-of-sight (i.e. forward AND return for a reflected channel: NO dividing by 2 correction)"

AreWeDreaming:
I have been trying to get TS and CO2 to agree in my IDA workflow for ages. They don't and a factor of 2 won't make things much better.

AreWeDreaming:
Also from the CO2 wiki entry: 
All signals are for the double pass line-integrated density along a given chord in (m/cm3) ≡ (106 m-2): denr0

AreWeDreaming:
Nvm. the third point was always there, I just didn't see it =/

PR #223, opened 2023-01-10
fix ec cocos

jmcclena:
@smithsp @orso82 With this fix, the toray regression script should pass for the `all_omas` OMFIT branch  https://github.com/gafusion/OMFIT-source/pull/6454

### DIFF
--- a/omas/omas_cocos.py
+++ b/omas/omas_cocos.py
@@ -338,8 +338,8 @@ _cocos_signals['divertors.divertor.:.current_incident.data']='?'                
 _cocos_signals['divertors.divertor.:.power_currents.data']='?'                                           #[ADD?]# 1.750000 # current  [W]
 
 # EC_LAUNCHERS
-_cocos_signals['ec_launchers.beam.:.launching_position.phi']='?'                 #[ADD?]# 2.000000 # phi  [rad]
-_cocos_signals['ec_launchers.beam.:.steering_angle_pol']='?'                     #[ADD?]# 2.000000 # _pol  [rad]
+_cocos_signals['ec_launchers.beam.:.launching_position.phi']='TOR'                      # 2.000000 # phi  [rad]
+_cocos_signals['ec_launchers.beam.:.steering_angle_pol']=None                           # 2.000000 # _pol  [rad]
 _cocos_signals['ec_launchers.beam.:.steering_angle_tor']='TOR'                          # 2.000000 # _tor  [rad]
 
 # ECE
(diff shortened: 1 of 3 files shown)


omas/omas_cocos.py
```diff
@@ -338,8 +338,8 @@
 _cocos_signals['divertors.divertor.:.power_currents.data']='?'                                           #[ADD?]# 1.750000 # current  [W]
 
 # EC_LAUNCHERS
-_cocos_signals['ec_launchers.beam.:.launching_position.phi']='?'                 #[ADD?]# 2.000000 # phi  [rad]
-_cocos_signals['ec_launchers.beam.:.steering_angle_pol']='?'                     #[ADD?]# 2.000000 # _pol  [rad]
+_cocos_signals['ec_launchers.beam.:.launching_position.phi']='TOR'                      # 2.000000 # phi  [rad]
+_cocos_signals['ec_launchers.beam.:.steering_angle_pol']=None                           # 2.000000 # _pol  [rad]
```
smithsp, line 342: I find this odd to have no COCOS dependence.

smithsp (commented):
@bclyons12 Can you check the COCCOS here?

omas/omas_cocos.py
jmcclena: hmm... yeah. Maybe it should be 'POL' now that you mention it.

orso82: @jmcclena the definition is: "Steering angle of the EC beam in the R,Z plane (from the -R axis towards the -Z axis), angle_pol=atan2(-k_Z,-k_R), where k_Z and k_R are the Z and R components of the mean wave vector in the EC beam". The COCOS definition for the poloidal angle is starting from the +R axis (and COCOS 11 has the polodal angle to be clockwise). This means that the `steering_angle_pol` definition in IMAS is not consistent with COCOS 11. I would just leave it as `None`.

jmcclena: So should we merge this then?

smithsp:
Actually, here is the error
```
Run actions/setup-python@v2
  with:
    python-version: 3.6
    token: ***
Version 3.6 was not found in the local cache
Error: Version 3.6 with arch x64 not found
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```
Maybe a fluke, as `3.6.15` and others are in the link.

orso82:
The regression failing is strange. The issue seems to be with the GitHub CI and I would be ok to merge as is

smithsp:
I see that is an option.  I am OK as well.  I commented on https://github.com/actions/setup-python/issues/585 to try to track GitHub CI problem.

Issue #213, opened 2022-10-05
omas install fails from a conda .yml file.


cianciosa:
Take the simple `.yml` file to define a Conda environment.

```
name: omas
dependencies:
  - pip
  - pip:
    - omas
```

When building this environment, the pip command to install omas fails when trying to install the pygacode dependency.
```
conda env create -f omas.yml --force
Collecting package metadata (repodata.json): done
Solving environment: done
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
Installing pip dependencies: / Ran pip subprocess with arguments:
['/global/homes/m/markcian/miniconda3/envs/omas/bin/python', '-m', 'pip', 'install', '-U', '-r', '/global/u2/m/markcian/ips-fastran/install/conda/condaenv.irdo0hc6.requirements.txt']
Pip subprocess output:
Collecting omas
  Using cached omas-0.87.0.tar.gz (30.4 MB)
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Collecting numpy>=1.16.1
  Using cached numpy-1.23.3-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (17.1 MB)
Collecting uncertainties
  Using cached uncertainties-3.1.7-py2.py3-none-any.whl (98 kB)
Collecting pint
  Using cached Pint-0.19.2.tar.gz (292 kB)
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Preparing metadata (pyproject.toml): started
  Preparing metadata (pyproject.toml): finished with status 'done'
Collecting netCDF4
  Using cached netCDF4-1.6.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (5.1 MB)
Collecting boto3
  Using cached boto3-1.24.87-py3-none-any.whl (132 kB)
Collecting matplotlib
  Using cached matplotlib-3.6.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (11.8 MB)
Collecting scipy
  Using cached scipy-1.9.1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (43.9 MB)
Collecting h5py
[...]

PR #232, opened 2023-02-21
Remove pygacode as a required dependency

LiamPattinson:
I've been trying to use some [OMAS functions in Pyrokinetics](https://github.com/pyro-kinetics/pyrokinetics/pull/161), and although everything worked fine on my machine, it caused our automated tests to fail. It looks like the same problem that was encountered in [this issue](https://github.com/gafusion/omas/issues/213), in which the build-time `numpy.distutils` requirement of pygacode was causing the install to fail.

As pygacode isn't used directly by OMAS (as far as I can tell), here I've simply removed it as a required dependency. It's still available as part of the `'machine'` extras, though I've also updated that to also not be a required install. To fix this issue in pygacode, somebody will simply need to add a `pyproject.toml`, as described in [this comment](https://github.com/gafusion/omas/issues/213#issuecomment-1437389680).

A concern I have is that these changes make `omfit_classes` an optional install, and although OMAS is written to permit this, other users may not be using `pip install omas[machine]` in their workflows and therefore this may break their automated systems. Should `omfit_classes` be moved from `extras_require.machine` to `install_requires`? As far as I can tell, neither `pexpect` nor `fortranformat` from `extras_require.machine` are currently in use. The simplest fix would be to move `omfit_classes` to `install_requires` and remove the other dependencies.

Is all of this okay? I'm not sure how OMAS is used by the majority of users, so I'm aware that I may be missing some context regarding the inclusion of pygacode as a dependency.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,7 +20,6 @@ Cython                   # required
 omfit_classes            # required
 pexpect                  # required
 fortranformat            # required
-pygacode                 # required
 
 # omfit_classes            # machine
 # pexpect                  # machine
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -34,7 +34,7 @@ jobs:
 
     - name: Install OMAS
       run: |
-        python3 -m pip install .
+        python3 -m pip install .[machine]
 
     - name: Move to root and try importing OMAS
       run: |
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
-        python-version: '3.6.15'
+        python-version: '3.6'
 
     - name: Install numpy
       run: |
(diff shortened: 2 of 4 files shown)


LiamPattinson:
I had a go at getting the automated tests to work, and while they are running now, I'm not familiar enough with the structure of the library to diagnose the problem here.

requirements.txt
```diff
@@ -20,7 +20,6 @@ Cython                   # required
 omfit_classes            # required
 pexpect                  # required
 fortranformat            # required
-pygacode                 # required
```
orso82, line 23: All these packages are only required to support machine mappings.
I can agree to make `machine` an extra and not a requirement.
Then we can remove `omfit_classes`, `pexpect`, `fortranformat` in addition to `pygacode`.

orso82:
@LiamPattinson thank you for fixing the OMAS regression tests! I have cherry-picked those changes and now testing not having `omfit_classes` `pexpect` `fortranformat` `pygacode` be required packages.

PR #236, opened 2023-03-08
Added profiles from CAKE

AreWeDreaming:
- Ripped out other references to profiles needs to be reviewed carefully


omas/machine_mappings/d3d.json
```diff
@@ -113,32 +113,39 @@
  "coils_non_axisymmetric.coil.:.name": {
   "PYTHON": "coils_non_axisymmetric_hardware(ods, {pulse})"
  },
- "core_profiles.global_quantities.v_loop": {
-  "COCOSIO": 11,
-  "PYTHON": "core_profiles_global_quantities_data(ods, {pulse})"
- },
```
AreWeDreaming, line 119: I had to rip this out to make it work. Putting it back in causes:
``core_profiles.profiles_1d.0.grid.rho_pol_norm` has no data`

omas/machine_mappings/d3d.json
```diff
@@ -113,32 +113,39 @@
...
  },
- "core_profiles.profiles_1d.:.electrons.density_thermal": {
-  "TDI": "\\{PROFILES_tree}::TOP.PROFILES.EDENSFIT",
-  "treename": "{PROFILES_tree}"
+ "core_profiles.profiles_1d.:.electrons.density": {
+  "PYTHON": "core_profiles_profile_1d(ods, {pulse})"
  },
  "core_profiles.profiles_1d.:.electrons.temperature": {
-  "TDI": "\\{PROFILES_tree}::TOP.PROFILES.ETEMPFIT",
-  "treename": "{PROFILES_tree}"
```
AreWeDreaming, line 130: Same here, but not because of errors but because of multiple sources for the same field. I don't know how to resolve this.

omas/machine_mappings/d3d.json
smithsp: @orso82 Do you understand why @AreWeDreaming would get this error, if this is not removed?

omas/machine_mappings/d3d.json
smithsp: @orso82 How does one specify in the machine mappings when the same quantity is available via a variety of MDSplus trees?  Maybe passing an argument to `ods.open` that gets passed to the underlying new `core_profiles_profile_1d`?

smithsp (commented):
Looks promising.

orso82:
See the "__options__" section in the "*.json" files to see how options can be passed to mapping functions. For example, you should be able to do:
```python
"__options__": {
  "EFIT_tree": "EFIT01",
  "PROFILES_tree": "ZIPFIT01",
  "analysis_type": "CERQUICK",
  "default_tree": "D3D",
  "fast_ece": false,
  "nref": 0,
  "revision": "BLESSED"
 },
...
"core_profiles.profiles_1d.:.electrons.density": {
  "PYTHON": "core_profiles_profile_1d(ods, {pulse}, {PROFILES_tree})"
 }
```

which can then be passed with
```python
with ods.open("d3d", 133221, PROFILES_tree="mytree")
    plot(ods['core_profiles.profiles_1d.0.electrons.density'])
```

orso82:
Should the ability to read ZIPFITS be retained? Are you sure there is nothing that depends on that?

omas/machine_mappings/d3d.json
orso82: Maybe `core_profiles_global_quantities_data ` looks at `core_profiles.profiles_1d.0.grid.rho_pol_norm` for some reason, and that mapping was not in place?

AreWeDreaming:
@orso82 Thanks for the example. I am not suggesting to removed anything, I was just confused about what to do and wanted to have an example for what I was struggling with when asking for help. I didn't realize that I need to move the zipfit mapping out of the .json file and into my new routine. I'll add an if-clause and the tree name as an argument.
The Omas viewer should be able to show whatever Omas can load from zip fits.

AreWeDreaming:
The last commit has several problems:
1. The "PROFILES_tree" extra argument has to include `\'`, e.g. `"\'ZIPFIT01\'"` in order to not loose its string nature. This is not the case for `"EFIT_tree"` which is a very similar extra argument.
2. `PROFILES_tree="\'OMFIT_PROFS\'"`and  `PROFILES_tree="\'ZIPFIT01\'"` do not provide the same fields. This seems to confuse the `.json` file that wants to have a fixed set of fields for each mapping function regardless of its input argument. 
3. I also need uncertainties. I learned from Sean that the `TDI` function is `error_of` but I am not sure how to do this for all fields without explicitly looping. I am not sure if explicitly looping over each field might cause separate MDS+ requests for each quantity. I guess I could just create a query like this:
```
query = {
            "electrons.density.data_error_upper": "error_of(\\TOP.n_e)",
            "electrons.temperature.data_error_upper": "error_of(\\TOP.T_e)",
            "ion[0].density.data_error_upper": "error_of(\\TOP.n_D)",
            "ion[0].temperature.data_error_upper": "error_of(\\TOP.T_i)",
            "ion[0].velocity.toroidal.data_error_upper": "error_of(\\TOP.V_tor_C)",
            "ion[0].density.data_error_upper": "error_of(\\TOP.n_C)"
        }
```

orso82:
1. see how it's done for `charge_exchange.channel` for example
2. I need to see the error, if there is an error
3. Indeed you'll have to query the uncertainties separately and then assign them to the ODS

AreWeDreaming:
Ok, so I managed to sort out all the problems I think. It seems that there were some minor errors that then became several weird problems. Now I can switch between `ZIPFIT`/`OMFIT_PROFS` without issues and the tree specification works without `\"`.
I also added the uncertainties.
We'll have to revisit this when the fields for `*_fit` have data in MDS+. For now, I am happy with this.

smithsp:
@AreWeDreaming should remove the `no-pr-activity` label if this PR is still valid.  If it needs further testing, let's work on it.  The last automatic test failed.

AreWeDreaming:
Now passing all checks

torrinba (approved):
This passes all (functional) OMFIT tests that I know are relevant https://github.com/gafusion/OMFIT-source/pull/7238

Since it includes the fix to work with newer matplotlib versions (which is preventing any other PRs from passing the tests), I think it should be merged immediately

PR #242, opened 2023-03-23
WIP: Refactored `omas_machine` to remove ring inclusion

AreWeDreaming:
- Moved functions needed by mappings to seprate file
  - Moved the machine mapping decorator to a utilities file
  - Moved MDS+ related routines into their own file
- Added extra argument to pass `update_mapping` as an argument
- Removed `compile` statement as it was causing trouble for the debugger
- Ran `test_machine_mapping_functions` for `d3d.py` and fixed bugs that appeared
- WIP state because `NSTX` and `EAST` machine mappings not tested,

### DIFF
--- a/omas/machine_mappings/d3d.py
+++ b/omas/machine_mappings/d3d.py
@@ -1028,6 +1028,8 @@ def langmuir_probes_data(ods, pulse, _get_measurements=True):
         if data_present[i]:
             try:
                 r = mdsvalue('d3d', pulse=pulse, treename='langmuir', TDI=r'\langmuir::top.probe_{:03d}.r'.format(i)).data()
+                if r is None:
+                    raise ValueError()
             except Exception:
                 continue
             if r > 0:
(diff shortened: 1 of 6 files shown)


omas/machine_mappings/d3d.py
```diff
@@ -1024,6 +1028,8 @@ def langmuir_probes_data(ods, pulse, _get_measurements=True):
         if data_present[i]:
             try:
                 r = mdsvalue('d3d', pulse=pulse, treename='langmuir', TDI=r'\langmuir::top.probe_{:03d}.r'.format(i)).data()
+                if r is None:
+                    raise ValueError()
```
AreWeDreaming, line 1032: Why does MDS+ return `None` for the test. This causes a comparison of `None > 0` later. I put in this hotfix, but I am not happy with it because I don't understand how the old code worked. Also why does the test case have no data? That seems a bit pointless.

smithsp:
Is the WIP label still appropriate here?

orso82:
Merge @AreWeDreaming ?

PR #245, opened 2023-04-17
Fix owner check in omas_utils.py

### LINKED ISSUE #240: Issues in docker and/or CI (env var USER)
Hi.

An error is raised during import when trying to use `omas` in a **docker container** and/or **gitlab CI** pipeline, where the default user is root and traditional configurations (`.bashrc`, `.profile`) might be a bit different. The issue is the `USER` env variable, which is not set in these environments. It works with this workaround:
```bash
export USER=''
python -c 'import omas'
```

### Proposed solution

The exception points to https://github.com/gafusion/omas/blob/master/omas/omas_utils.py#L869

I feel like this check could be replaced by a simpler one using only the `UID` (which is set in docker):
```python
if os.name == 'nt' or os.environ['UID'] != os.stat(__file__).st_uid:
```
Although I have not gurantee that this is more correct and universal (will try to google), it feels like it could be.

Or just to be sure, using `get()` would also help
 ```python
... or os.environ.get('UID') != ...
```

hetsko:
Closes #240 

I have replaced the file ownership check via the `USER` env variable (which is not guaranteed to be set, e.g. when running as root in docker) with a call to `os.geteuid()` to check against **effective uid** of the process.

**euid** vs. **uid**: I am not an os expert, but I am quite confident that checking against euid (instead of uid) is also the correct thing to do here, since I believe the file read/write system calls also operate with euid of the current process. But that is just an edge case, usually uid==euid.

### DIFF
--- a/omas/omas_utils.py
+++ b/omas/omas_utils.py
@@ -864,9 +864,9 @@ def omas_global_quantities(imas_version=omas_rcparams['default_imas_version']):
     return _global_quantities[imas_version]
 
 
-# only attempt cython if user owns this copy of omas
+# only attempt cython if effective user owns this copy of omas
 # disabled for Windows: need to add check for file ownership under Windows
-if os.name == 'nt' or os.environ['USER'] != pwd.getpwuid(os.stat(__file__).st_uid).pw_name:
+if os.name == 'nt' or os.geteuid() != os.stat(__file__).st_uid:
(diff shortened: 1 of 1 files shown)


orso82:
Thank you for your contribution @hetsko and sorry for getting only now on this!

PR #247, opened 2023-04-20
Updating the mhdin.dat files to new EFIT/EFUND format

bechtt:
This is being done ahead of the OMFIT changes and updates to EFIT system installs.

It is required as the newest version of EFIT eliminates dprobe.dat files and therefore needed to make changes to mhdin.dat


jmcclena:
@bechtt It appears the NSTX 133139 FCID is still off in this pull request. I thought I fixed this on the EFIT side. Maybe I forgot to push it.

Screen Shot 2023-06-21 at 9 57 59 AM

bechtt:
> The changes look ok to me. I don't know how to test them. Does @bechtt have permission to merge here?

I do not have merge permission for omas.

Joey and I have been testing this out which has uncovered a handful of unrelated bugs that have hopefully all been fixed at this point.  When he gives the thumbs up we will be ready to merge.

We do face a bit of a challenge in the fact that this is directly related to changes in EFIT and also an associated OMFIT PR https://github.com/gafusion/OMFIT-source/pull/6607.  We don't expect the OMFIT regression tests to pass until both this PR and the EFIT system installation is updated.  What's the best way for us to line all of that up without too much disruption?

smithsp:
The `omas` git submodule in https://github.com/gafusion/OMFIT-source/pull/6607 should point to the latest commit of this branch.  Then the public versions should be updated*, and when those are in place, then merge both omas and OMFIT.

* It may be useful to create a new module version that gets changed to the default for more seamless transition.

jmcclena (approved):

PR #248, opened 2023-05-17
Fixed bug that occurs for ods with no time entries

AreWeDreaming:
Small change that fixes a crash when trying to write an IDS from data that does not have any time (e.g. wall_ids)

### DIFF
--- a/omas/omas_imas.py
+++ b/omas/omas_imas.py
@@ -297,6 +297,8 @@ def save_omas_imas(ods, user=None, machine=None, pulse=None, run=None, occurrenc
 
     :param verbose: whether the process should be verbose
 
+    :param backend: Which backend to use, can be one of MDSPLUS, ASCII, HDF5, MEMORY, UDA, NO
+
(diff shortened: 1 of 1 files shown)


omas/omas_imas.py
```diff
@@ -274,7 +274,8 @@ def imas_get(ids, path, skip_missing_nodes=False, check_empty=True):
 # save and load OMAS to IMAS
 # --------------------------------------------
 @codeparams_xml_save
-def save_omas_imas(ods, user=None, machine=None, pulse=None, run=None, occurrence={}, new=False, imas_version=None, verbose=True):
+def save_omas_imas(ods, user=None, machine=None, pulse=None, run=None, occurrence={}, 
+                   new=False, imas_version=None, verbose=True, backend='MDSPLUS'):
```
smithsp, line 278: Since you added a keyword, you should modify the docstring.

AreWeDreaming: Fixed in new commit

PR #249, opened 2023-05-20
allow specify IMAS backend

orso82:
@jmcclena would you mind testing this branch next time you find yourself saving/loading data in IMAS


orso82:
I see that this was actually implemented in https://github.com/gafusion/omas/pull/248

PR #251, opened 2023-06-06
Apply COCOs transforms to the ..._error_upper quantities

bechtt:
Thanks to Orso for suggesting the generic way to handle this!

### DIFF
--- a/omas/omas_core.py
+++ b/omas/omas_core.py
@@ -931,20 +931,20 @@ class ODS(MutableMapping):
                     self.cocosio
                     and self.cocosio != self.cocos
                     and '.' in location
-                    and ulocation in omas_physics.cocos_signals
+                    and re.sub(r'_error_upper$', '', ulocation) in omas_physics.cocos_signals
                     and not isinstance(value, ODS)
                 ):
-                    transform = omas_physics.cocos_signals[ulocation]
+                    transform = omas_physics.cocos_signals[re.sub(r'_error_upper$', '', ulocation)]
                     if isinstance(transform, list):
                         norm = np.ones(len(transform))
                         for itf, tf in enumerate(transform):
                             norm[itf] = omas_physics.cocos_transform(self.cocosio, self.cocos)[tf]
                     elif transform == '?':
                         if isinstance(self.consistency_check, str) and 'warn' in self.consistency_check:
-                            printe('COCOS translation has not been setup: %s' % ulocation)
+                            printe('COCOS translation has not been setup: %s' % re.sub(r'_error_upper$', '', ulocation))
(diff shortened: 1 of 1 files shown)


smithsp (commented):
My one concern is that you might be having a sign conversion for the base value, and then error_upper shouldn't get the sign conversion.

omas/omas_core.py
```diff
@@ -931,20 +931,20 @@ def __setitem__(self, key, value):
                     self.cocosio
                     and self.cocosio != self.cocos
                     and '.' in location
-                    and ulocation in omas_physics.cocos_signals
+                    and re.sub(r'_error_upper$', '', ulocation) in omas_physics.cocos_signals
```
orso82, line 934: @bechtt @smithsp to improve performance one suggestion would be to store `re.sub(r'_error_upper$', '', ulocation)` in a separate variable. You could also use `re.sub(r'_error_(upper|lower)$', '', ulocation)` to make the regex also work for `_error_lower`.

bechtt: Yeah, I can make that change.  Does `blocation` or `vlocation` seem like a logical choice?

@orso82 is there a nicer way to ensure the uncertainties remain positive definite instead of adding another conditional statement inside of these definitions?

bechtt: > what do you mean by `positive definite` ?

Following up on Sterling's concern https://github.com/gafusion/omas/pull/251#pullrequestreview-1465946992.  Maybe I am misunderstanding.  

Does a value of -5 and uncertainty of -1 make sense or should the uncertainty be 1, assuming there was a COCOs sign conversion, for example?

orso82: The standard deviation is a positive quantity, so I think that `..._error_upper` should always be positive.

orso82:
Great contribution @bechtt and good thinking @smithsp ! If you are happy with this I think we can merge.

smithsp (approved):

smithsp:
@bechtt How did you test this?  Maybe good to add a test to the test suite.  But I am also OK to just merge.

bechtt:
> @bechtt How did you test this? Maybe good to add a test to the test suite. But I am also OK to just merge.

I used OMFIT to convert eqdsk files (specifically the flux loops have a COCOs transformation and uncertainty).  That doesn't seem like the best candidate for an omas test unfortunately.  I also did not test the `...error_lower` conversion or the impact of a sign flip in the COCOs.

orso82:
@bechtt if you feel inspired, you can add a regression here:
https://github.com/gafusion/omas/blob/master/omas/tests/test_omas_core.py#L103-L107

PR #252, opened 2023-06-06
Matplot latest

wdeshazer:
Changes that make omas compatible with matplot_lib 3.5 have been incorporated.


smithsp:
@wdeshazer When I ran
```
cd omas
pip install -e .
python3  omas/tests/test_omas_plot.py
```
I was able to see that there was a missing import.  See https://github.com/gafusion/omas/pull/252/commits/401f695f214fcf4158b8a141c58c62d4f583710a

PR #255, opened 2023-06-29
Fill out more of the EFIT mapping to MDS+

bechtt:
This may not be 100% complete, but should be pretty close.

PR #258, opened 2023-07-31
Omas plot w new efit mapping

AreWeDreaming:
Dependent on #256. This PR uses the new fields from #255  for the equilibrium summary plot. Regression test for `omas_plot` passes locally.


AreWeDreaming:
This really should be merged. It has some pretty sizeable changes. Notably #255 needs to be merged first.

AreWeDreaming:
Oh #255 already was merged so that is good.

torrinba (approved):
This all looks fine to me. And passes the OMAS and OMFIT regression tests.

Once you fix the conflicts I think we should merge @AreWeDreaming

PR #267, opened 2023-08-23
The E coils and F coils do not use same unit

AreWeDreaming:
The E and F coils are combined when loading from the EFIT tree. However, the F coils have the turns multiplied with the current, which one could argue is an preexisting error in MDS+. This would be at least consistently wrong but the F-coils are combined with the E-coils in `pf_current` in `constraints` which are stored correctly.
Following options:
- Merge this and remove the mapping completely
- Try to fix the mapping, but this is tricky because the information on how many windings each coil has is not MDS+. For DIII-D there is some expectation that the F-coils always have the same amount of turns for all shots, but the EFIT tree and mapping is not DIII-D specific, so it gets complicated very quickly.
- Leave it because it is consistent how the code wants its `pf_current`s. Here it is important to remember that this is the mapping for EFIT not the mapping for `pf_active`.
I leave it to @smithsp to make an executive decision and @bechtt to add his own opinion.

### DIFF
--- a/omas/machine_mappings/_efit.json
+++ b/omas/machine_mappings/_efit.json
@@ -315,20 +315,8 @@
    "TDI": "size(\\{EFIT_tree}::TOP.MEASUREMENTS.ECCURT,0) + size(\\{EFIT_tree}::TOP.MEASUREMENTS.FCCURT,0)",
    "treename": "{EFIT_tree}"
  },
- "equilibrium.time_slice.:.constraints.pf_current.:.measured": {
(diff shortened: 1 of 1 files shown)


bechtt:
The second option would be the best in my opinion.  The MDS+ organization used by different machines (if they even do) is not universal, so I would consider most of the EFIT mapping to be DIII-D specific.  Some parts are the same for NSTX (and are used), but I would consider that an exception rather than the norm.

I also think leaving `pf_current` with different units/interpretations for different elements is confusing for any use case besides EFIT and does not follow the IMAS definition (`pf_current` specifies it should have the same size as `pf_active.coil` and `pf_active.coil[:].current` is defined as the measured current in a single turn).

My second choice would be the first option, which should be fine temporarily since this isn't being used anywhere yet.

PR #277, opened 2023-10-09
`machine_to_omas` use temp cache for wildcard load

AreWeDreaming:
Small change to always use caching for wildcard loads. Big speed improvement because now each python mapping function is only used once per wildcard load.
Leaf of #258 and should be merged accordingly.


AreWeDreaming:
I need to get better at keeping merging my branches in the correct order. Oh well at least it is merged.

smithsp:
@AreWeDreaming This is the wrong one to close, I think.

AreWeDreaming:
You're right.

PR #283, opened 2023-12-12
Mapping to/from rho tor norm

AreWeDreaming:
Leaf of #258
Adds `rho_tor_norm` to the magnetic coordinates `omas_physics.remap_flux_coordinates` can map between. Wouldn't hurt if this someone double checked my math once #258 is merged. I did do a spot check and while the result wasn't perfect it was not too far off.
Arguably a unit test wouldn't hurt here, but I don't have time to write it at the moment.


torrinba (approved):
I see no new changes here, but if you can point me to where the `rho_tor_norm` conversion was added I'm happy to look it over

AreWeDreaming:
https://github.com/gafusion/omas/blob/6f2afabbab210cc9285b55b6720a6f5176e688c6/omas/omas_physics.py#L390
and above

torrinba:
Is this comment still accurate? It appears not to be...
```
def remap_flux_coordinates(ods, time_index, origin, destination, values):
    """
    Maps from one magnetic coordinate system to another. At the moment only supports
    psi <-> rho_pol
```

torrinba:
Have you tested whether the transformations to/from `phi` are correct? If it isn't already in the ODS, I have some concerns about this working correctly, without accounting for the boundary terms
```
phi_spline = q_spline.antiderivative(1)
```
Of course the magnitude of `phi` doesn't matter if you're just interested in `rho_tor_norm`

torrinba:
https://github.com/gafusion/omas/blob/6f2afabbab210cc9285b55b6720a6f5176e688c6/omas/omas_physics.py#L313

This line assumes that the data already exists, which might not be the case if the conversion is from `rho_tor_norm`.  It would be more robust to protect against the `ValueError` in this case by falling back on `add_phi_to_equilbrium_profiles_1d_ods` as well

torrinba:
```
phi_spl = InterpolatedUnivariateSpline(psi_grid[psi_grid_mask]
```
Assumes that `psi` is strictly increasing (AFAIK this is a requirement for `InterpolatedUnivariateSpline`), but the sign of `psi` depends on the COCOs so I don't think this is a safe assumption.

smithsp:
This PR still shows no changes relative to `master` branch.

AreWeDreaming:
Well we should probably address the lack of cocoa awareness when dealing with phi. Also a unit test would be in order, but I guess yes an issue would be more appropriate

PR #290, opened 2024-02-15
fix(structure): ensure add_extra_structures resets all required caches

hassec:
I think there is a small bug in `add_extra_structures`

If the structure is already loaded and then extended the extension doesn't properly show up. 
See below, where `test_field` is missing from the output of the second `imas_structure` call.
image

This means that e.g. a `ODS().load()` call will fail because the structure caches for `limiter.unit` will not contain `test_filed`.

This PR adds `omas.omas_utils._ods_structure_cache = {}` to the other resets in `add_extra_structures` to fix this behaviour.

I've also added a test for this.

### DIFF
--- a/omas/omas_structure.py
+++ b/omas/omas_structure.py
@@ -588,6 +588,7 @@ def add_extra_structures(extra_structures, lifecycle_status='tmp'):
     # reset structure caches
     omas_utils._structures = {}
     omas_utils._structures_dict = {}
+    omas_utils._ods_structure_cache = {}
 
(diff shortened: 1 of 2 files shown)


orso82:
fantastic!

hassec:
Thanks for the instant reply @orso82 😊 
Any chance we can get a new PyPi release soon? There have been a couple of things merged that we are looking forward too.

orso82:
done. Version 0.93.0 was released

Issue #291, opened 2024-02-22
test_machines Is Failing on master


torrinba:
For some reason it looks like the svn setup with the machine mappings is now failing.  I didn't see this problem last time I ran the tests and believe everything I am doing is the same (this was run with `make test` on Iris using the omfit/unstable environment).  

```
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
TestOmasMachine.test_machines
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
svn: 'https://github.com/gafusion/omas.git/trunk/omas/machine_mappings' path not found
svn: 'https://github.com/gafusion/omas.git/branches/dummy/omas/machine_mappings' path not found
svn: 'https://github.com/gafusion/omas.git/trunk/omas/machine_mappings' path not found
ERROR

======================================================================
ERROR: test_machines (test_omas_machine.TestOmasMachine)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "omas/omas/tests/test_omas_machine.py", line 47, in test_machines
    assert omas_rcparams['tmp_omas_dir'] in machines(self.machine, 'master')
  File "omas/omas/omas_machine.py", line 548, in machines
    raise NotImplementedError(f'Machine mapping file `{machine}.json` does not exist')
NotImplementedError: Machine mapping file `d3d.json` does not exist

----------------------------------------------------------------------
Ran 64 tests in 143.150s

FAILED (errors=1, skipped=10)
make: *** [test] Error 1
```

There's a similar test in `test_remote_machine_mappings` which will also fail if this one is bypassed.  Does anyone have an idea why this wouldn't work?  How important is this?

AreWeDreaming:
The `d3d.json` stores all the mappings of DIII-D data to IMAS (except for efit). So it is quite important. That being said I think it is completely autogenerated, so I am not sure why it doesn't appear on its own. This is my limited understanding, @orso82 would know more.

torrinba:
The tests that use `d3d.json` from the local omas version also pass just fine.  It's only when omas tries to fetch it from master that the tests fail.  I don't understand the use case for this either

PR #292, opened 2024-02-22
Update to use IMAS 3.40.1

torrinba:
This seems like a pretty small change overall.  I haven't seen much change in the DD and all tests are still passing (both here an for OMFIT).

The PR for updating the OMAS version in OMFIT should be merged after this


torrinba:
Is there a way to look at why the regression tests failed here?  It could be related to this https://github.com/gafusion/omas/issues/291 but I don't see how to confirm this.  All other tests passed when I ran them on Iris so I can deactivate those tests for now if the fixes will be longer term (I can see the last 2 merges did not pass the regression)

PR #293, opened 2024-02-27
refactor(build): change setup.py to pyproject.toml and supress cython warning

### LINKED ISSUE #289: Suppress spurious compiler warnings on first import of omas
On first import of `omas`, the compilation of the cython extensions will dump a ton of warnings. 
See e.g.: 
image

And this continues for O(100) lines. IMHO this is quite confusing for new users.

It would be nice if there was a way to suppress these or even better build extensions as part of the packaging to in addition to avoiding these warnings, one also avoids the slowdown on the first import.

hassec:
I was motivated to fix #289, but when starting to work on the old `setup.py` I was curious if you would be open to a bit of a modernisation, see below:

Changes in this PR: 

- Move package/build configuration into a `pyproject.toml` following the adopted python standards.
- Only have a very minimal `setup.py` that translates `omas_cython.pyx` to `omas_cython.c` at the time the sdist package is created. This has the benefit that `Cython` is no longer a dependency for users. They `*.c` file still gets compiled during package install on the users computer, but that still is pretty quick and doesn't produce any warnings.

This PR also removes the need to manually traverse the file hierarchy to create `packages` and `package_data` lists. This is now automatically handled.

Editable installs still work by simply running:
```
pip install -e .
```

To build a sdist package using [`build`](https://github.com/pypa/build) run:
```
python -m build --sdist
```
from the root of the project. This creates an isolated virtual environment to build the sdist according to the spec in `pyproject.toml`.


omas/omas_utils.py
```diff
@@ -864,23 +865,6 @@ def omas_global_quantities(imas_version=omas_rcparams['default_imas_version']):
     return _global_quantities[imas_version]
 
 
-# only attempt cython if effective user owns this copy of omas
-# disabled for Windows: need to add check for file ownership under Windows
-if os.name == 'nt' or os.geteuid() != os.stat(__file__).st_uid:
```
hassec, line 869: With this change the compiled version of `omas_cython` should be available to everyone. Also windows users.

pyproject.toml
```diff
@@ -0,0 +1,63 @@
+[build-system]
+requires = ["setuptools", "Cython"]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools.packages.find]
+where = ["."]
+exclude = ["sphinx*"]
+
+[tool.setuptools.dynamic]
+version = {file = "omas/version"}
```
hassec, line 10: This automatically sets the version to what is found in `omas/version`

orso82:
We're certainly very open!!! Thank you @hassec for taking a deep dive into the setup files of OMAS!!

Perhaps @kalling can comment here. I have not kept up to date with the latest and greatest ways to setup a Python package.

pyproject.toml
```diff
@@ -0,0 +1,68 @@
...
+classifiers = [
+    'Development Status :: 5 - Production/Stable',
+    'Programming Language :: Python',
+    'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3 :: Only',
+    'License :: OSI Approved :: MIT License',
+]
+keywords = ["integrated modeling",  "OMFIT", "IMAS",  "ITER"]
+
+requires-python = '>=3.8'
```
hassec, line 33: It would require bumping this from 3.6 to 3.8. 
3.6 was end of life in 2021, so I hope that this is ok?

smithsp: OMFIT is at python=3.7 and uses omas as an integral part.  What changes are you making here that require python>=3.8?

hassec: I looked at what the recent versions of setuptools require.
But what would be needed here is support for [PEP-621](https://peps.python.org/pep-0621/) which looks like it landed in `setuptools` version [v61-0-0](https://setuptools.pypa.io/en/latest/history.html#v61-0-0) which still supports >=3.7. 

I could try and see giving if it works for 3.7.

hassec: @smithsp I double checked and I was correct that in theory `3.7` should have been sufficient to enable usage of `pyproject.toml` based package configurations.

However, the problem here is the dependency on `pygacode` which is forced to version `0.71` from 2022 because that's the last version that supports `3.7`, and that version does not correctly specify its build dependencies (requires numpy). This is fixed in for versions `>1.0` but those require python `>=3.8` which is why the build here works for `3.8` but nor for `3.7`.

smithsp: We are actively working on migrating OMFIT to a later python version.

smithsp:
I don't think these commits got merged, unfortunately.

pyproject.toml
hassec: @smithsp This was the blocker why the PR wasn't fully ready to get merged. Is Python 3.7 still required to be supported?

smithsp: @hassec Thanks for following up.  Yes, python 3.7 is still required to be supported for OMFIT.  We looked into migrating OMFIT to python >=3.8 and there were problems/changes with the multiprocessing library of python that we have not worked around yet.

hassec: Thanks for the update @smithsp. Once the move to 3.8 is done, updating this PR should be pretty straight forward, and I'd be happy to do that. 

But I'm afraid that I likely won't be able to find the time in the next few weeks to dig into possible workarounds to make this PR work with 3.7.

smithsp: No worries.  This PR got closed automatically, and I just re-opened it to make sure we didn't lose these changes.

smithsp:
@torrinba @AreWeDreaming I found this as an open browser tab; I had meant to follow up on it.

PR #294, opened 2024-02-27
Upgrade python version to 3.8

hassec:
During debugging of the pipeline failure in #293, I noticed that the errors actually seem to come from the upgrade to python 3.8.

This PR currently only upgrades the python version to 3.8, which is a prerequisite for #293.

Before I spent more time on this I'd like to understand if an upgrade of the python version is an option for you. 3.6 is by now quite old and was end of life in 2021, and even 3.8 is going to be end of life this October, [see here](https://devguide.python.org/versions/)

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
-        python-version: '3.6'
+        python-version: '3.8'
 
     - name: Install numpy
       run: |
--- a/omas/omas_plot.py
+++ b/omas/omas_plot.py
@@ -800,7 +800,7 @@ def equilibrium_CX(
             z = scipy.ndimage.zoom(z, sf)
             value_2d = scipy.ndimage.zoom(value_2d, sf)
 
-        cs = ax.contour(r, z, value_2d, levels, **kw)
+        cs = ax.contour(r, z, value_2d, levels=levels, **kw)
 
         if label_contours or ((label_contours is None) and (contour_quantity == 'q')):
             ax.clabel(cs)


orso82:
I'd like @smithsp @kalling to comment on this. In general I think it's better to be compatible with more versions of Python, rather than fewer, unless of course there is some gain from upgrading.

@hassec what do you think is the benefit of making OMAS compatible with Python 3.8 and up, instead of 3.6 and up?

hassec:
@orso82 in general I agree that there is no harm in supporting older python versions. But it is a tradeoff as it usually also means one can't use newer features like. 

Requiring python >= 3.8 happened in matplotlib 2 years ago and for recent matplotlib versions it's already >= 3.9. Similar things are true for numpy which is currently >= 3.9 and went to >=3.8 in August of 2021. 

I'd say going to `>= 3.9` would be a nice step, and then we could even consider fixing the tests and running the CI for `3.9, 3.10, 3.11, and 3.12` to also test for all of these versions.

kalling:
Given their tight coupling, until OMFIT uses `>=3.8` we probably don't want to restrict `omas` to `>=3.8`
OMFIT is currently `3.7`
Updating to newer versions of Python for `OMFIT` is in progress. A conda package with partial support for 3.8 was put on the omfit channel years ago as a test, but there are certain dependencies (Like MDSplus) that don't currently have an appropriate newer conda package nor an easy way to build them.  After we have a good process for those errant dependencies, keeping up with current versions of Python in the future should be much more straight forward.

smithsp:
Per https://anaconda.org/conda-forge/mdsplus/files we have MDSplus for various versions of python.

smithsp:
@kalling Good catch.  I need to go back and see why it was failing on osx, to see if it has been resolved with recent changes to the MDSplus build process.

PR #296, opened 2024-03-03
feat(ci): Run tests for more python versions

hassec:
I wanted to disentangle the fixing of builds for specific python versions from #293.

So this PR keeps the `setup.py` way of building OMAS and does not yet fix anything in regard to the Cython warnings. 

All that this PR does is update the CI to also run the tests for python versions 3.6-3.11,
and apply some small fixes or workarounds to make the OMAS tests pass for these python versions.

We only use OMAS with python >=3.10 so would love to see these tests as part of the CI to ensure everything works as expected on these versions too.


omas/tests/failed_imports.py
```diff
@@ -69,6 +69,8 @@
         failed_OMFIT = False
     except ImportError as _excp:
         failed_OMFIT = _excp
+    except AttributeError as _excp:
+        failed_OMFIT = "omfit_classes, from xarray import * bug "
```
hassec, line 73: `omfit_classes` runs `from xarray import *` which is broken on various `xarray` versions.
I've reported & fixed this upstream, [see here](https://github.com/pydata/xarray/pull/8791). So with the next release of `xarray` this should be fixed again.

orso82:
Thank you @hassec !! Fantastic! @smithsp @kalling can you please take a look as well? Then I think we can merge!

omas/tests/test_omas_examples.py
```diff
@@ -97,6 +97,12 @@ def test_uncertain(self):
 
     @unittest.skipIf(failed_OMFIT, str(failed_OMFIT))
     def test_plot_g_s_2_ip(self):
+        # on 3.7 this test raises:
+        # ValueError: Number of rows must be a positive integer, not 7.0
+        # It seems like this is caused by a bug in omfit_classes where a float
+        # instead of int is passed to plot
```
kalling, line 103: @orso82 Any idea if this has been fixed on the OMFIT side?  I assume we should just cast it?

kalling (approved):
Looks good as far as I can tell.

omas/tests/test_omas_examples.py
orso82: Probably this needs to be fixed on the OMFIT side. Could you please take care of it @kalling ?

kalling: @orso82 Alright.
@hassec Would you mind sending me (or pointing me to) a trace of the error that occurs if you don't skip this test?

hassec: @kalling no problem. It's in `omfit_gapy.py` which has the following lines:
```
        cplot = max([np.floor(np.sqrt(nplot)), 1.0])
        rplot = np.ceil(nplot / cplot)
```
which should probably we wrapped in `int(...)` 

Corresponding ci run with the error [is here](https://github.com/hassec/omas/actions/runs/8119419504/job/22195279200#step:10:948)

kalling: `cplot` is used in other calculations later that look like they may need it to be a float, so I submitted a change to `auto_merge` that stores the versions passed in to subplot as `plot_cols`  and `plot_rows` instead of doing the cast at assignment, just to be safe.  Re-ran the tests locally and it took care of the issue.

smithsp (approved):
Please double check that all of the desired tests run for all of the python versions requested.

hassec:
@smithsp was that comment for me? What would you like to see checked aside from what's shown in the CI?

omas/tests/test_omas_examples.py
hassec: @kalling @orso82 What's your preference here? Keep this for now or remove and wait for a new `omfit_classes` release with the fix?

kalling:
I verified the tests work locally with those `OMFIT` changes.  Testing here would probably need a new `omfit_classes` release as @hassec mentioned as a comment in my review.

omas/tests/test_omas_examples.py
kalling: Unless it's urgent that this get through, I'd say wait.

hassec: @kalling  What's the timeline to get a new `omfit_classes` release onto pypi?

kalling: Assuming the tests pass, it is a weekly process.

smithsp:
@hassec regarding https://github.com/gafusion/omas/pull/296#issuecomment-1979436546 , yes, I am just asking that you double check the CI outputs to be sure the 3.7 tests run the same tests as the other versions.

hassec:
@smithsp yes, on all versions we collect 141 tests. 
3.6 skips 13
3.7 skips 14 ( this is the `omfit_classes` bug we will wait for and as soon as that is on pypi we can remove the skipping of that additional test)
>= 3.8 skips 15 tests due to omfit + xarray problem.

omas/tests/test_omas_examples.py
hassec: @kalling are you sure? The [latest version on pypi](https://pypi.org/project/omfit-classes/#history) is from Nov 2023

hassec: @kalling @orso82 How would you feel about merging as is, and I can follow up as soon as the fix on `omfit_classes` is released on PyPi? Or if it's coming in the next days I'm also happy to wait. 

There isn't really a strong urgency, but I'd like to avoid that the PR sits too long, goes stale and gets forgotten.

kalling: @hassec Interesting, I wonder if the 2 factor auth introduced messed up the automation for uploading to `pypi`. It should currently be at `3.2024.09`
I'll take a look.

kalling: Ok, I fixed the `pypi` uploading issue for `omfit-classes` and `3.2024.9.2` was uploaded. Hopefully now it'll actually be weekly again :) (Assuming tests pass)
Thanks for pointing that out.

hassec: Thanks @kalling I'll have an update PR for OMAS ready asap to remove the workaround :)

PR #297, opened 2024-03-11
refactor(test) remove workaround for old omfit_classes version

hassec:
@kalling As promised, removing the workaround now that `omfit_classes` is updated

### DIFF
--- a/omas/tests/test_omas_examples.py
+++ b/omas/tests/test_omas_examples.py
@@ -97,12 +97,6 @@ class TestOmasExamples(UnittestCaseOmas):
 
     @unittest.skipIf(failed_OMFIT, str(failed_OMFIT))
     def test_plot_g_s_2_ip(self):
-        # on 3.7 this test raises:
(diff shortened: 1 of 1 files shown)


hassec:
Interesting, it installed `omfit_classes==3.2024.9.2` but failed. Will investigate

kalling:
@hassec It looks like the new master wasn't created yet. I am guessing some (OMFIT) tests failed, but will take a look. (The new package has the previous master, just much newer one than the last package posted to pypi)

kalling:
(I am basically saying the fixed code isn't in the pypi package yet)

hassec:
@kalling Where would I be able to see/check if a new `omfit_classes` is deployed? (to avoid bugging you here 😅)

kalling:
`omfit_classes` got updated on pypi yesterday due to the new master release, so theoretically the fix should be in.

hassec:
@kalling yep just reran the tests and they now all pass 🎉

PR #300, opened 2024-03-21
Setup new COCOs combination for em_coupling matrices

### LINKED ISSUE #299: OMAS COCOs transformations do not appear to be adequate for em_coupling
I'm not too familiar with what options that are available for COCOs transformations in OMAS, but from what I can tell they do not appear to be capable of supporting matrices which need to have different transformations applied for each dimension.

For example, the `mutual_plasma_active` matrix has units of inverse toroidal current (`TOR`) in one dimension (corresponding to pf_active coils) and units of flux (`PSI`) in the other dimension (corresponding to the poloidal flux generated by the coils).

This is a pretty unusual application for COCOs since it seems like the matrix would need to have the transformation `PSI/TOR` (which would be the same as `PSI*TOR` if `TOR` only involves a sign change), but I don't see any way to handle this in OMAS.  Is this possible?  

Currently `em_coupling` has no COCOs defined so any use of these variables from OMAS with non-native COCOs will produce incorrect results.  

The other `em_coupling` matrices seem like they would have the same transformation or something trivial (`TOR` or `None`)

torrinba:
The COCOs have now been defined for all of the different `em_coupling` matrices and should close #299.

I'm interested in why there is a distinction between `sigma_Ip` and `sigma_B0` since it would simplify things if we could eliminate one instead of multiplying them several places.  Can someone more familiar explain this?

### DIFF
--- a/omas/omas_cocos.py
+++ b/omas/omas_cocos.py
@@ -30,6 +30,7 @@
 #        transforms['F'] = transforms['TOR']
 #        transforms['POL'] = sigma_B0_eff * sigma_rhotp_eff
 #        transforms['BP'] = transforms['POL']
+#        transforms['PSI/TOR'] = sigma_Ip_eff * sigma_B0_eff * sigma_Bp_eff * (2 * numpy.pi) ** exp_Bp_eff
 #        transforms[None] = 1
 #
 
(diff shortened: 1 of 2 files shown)


orso82:
I honestly don't know or remember why we have both `sigma_Ip` and `sigma_B0`

jmcclena:
sigma_B0 exists in Sauter's paper (https://www.sciencedirect.com/science/article/pii/S0010465512002962). So I'd be surprised if one can be removed.

bclyons12:
`sigma_Ip = sign(Ip)` and `sigma_B0 = sign(Bt)`. Those don't have to be the same, so you need to maintain both. In DIII-D, for example, they are opposite signs in the standard configuration.

bclyons12:
1. Okay, I see what you mean. It's actually `sigma_Ip_eff` and `sigma_B0_eff`, which are how those quantities transform. Those should be the same, but I guess the COCOS paper leaves open the option that your transformation might want to enforce/change a sign for one of those:
image

2. `'PSI/TOR'` is supposed to be the transformation for inductance? I don't think `sigma_B0_eff` should come into that since the toroidal field isn't involved. Inductance is something like poloidal flux per unit current, so maybe you just want:
`transforms['PSI/TOR'] = sigma_Bp_eff * (2 * numpy.pi) ** exp_Bp_eff`?
Or maybe you don't want a transformation at all. Are you trying to maintain the mutual inductance as the translation from current/current change to poloidal flux/voltage. If so, I don't think it has a COCOS. It's just a number. If you want the mutual inductance to convert between **psi** and current, then this transformation is correct because the physical meaning of psi would change depending on the COCOS.

torrinba:
Thanks for the explanation Brendan!  If I'm understanding correctly, the `sigma_B0_eff` vs `sigma_Ip_eff` distinction is retained in order to support transformations outside the standard COCOs 1-18.  Does OMAS have options to specify a different transformation (such as only transforming B0 or Ip)?  It seems to me that a `cocosin/cocosout` index has to be specified for the transformations to be applied.

I'm having a hard time understanding how the mutual inductance would convey anything meaningful if it's convention didn't change when the definitions of flux and current are transformed.  I could be missing something there.  My use case for the `em_coupling` IDS is to use the mutual inductaces to compute the impacts of current sources.  So the transformation is important, but I can't say that is universal.  Whether the `sigma_B0_eff` and `sigma_Ip_eff` contributions are included has no consequence since I plan to stick with COCOs 1-18.

Orso also raised the possibility that the mutual inductance could have a mix of different COCOs if the flux and current each use a different COCOs, but we agreed this would be an edge case that should be handled on it's own and doesn't need to have dedicated support in OMAS.

bclyons12:
If you have more information on the definition of these matrices or how they get used, I might be able to advise.

torrinba:
The only official information I know of is the description provided by IMAS
https://gafusion.github.io/omas/schema/schema_em%20coupling.html

In practice we are storing the EFIT Green function tables there and using them in similar ways.  Joey has a paper being reviewed that is an example of one of the applications outside of EFIT
http://nike.gat.com/app/publications/tracker/item/16662

bclyons12:
How do the EFIT Green's function tables get used? You multiply those values by the coil current and it gives you the EFIT psi? If that's true, then yes, it has a COCOS definition because the EFIT psi is the poloidal flux per radian.

The easiest thing to do would be to write down the equation that has this mutual inductance in it. Then you can write down the transform for each part of the equation and see what's left over to make it balance.

torrinba:
FYI IMAS 4.0 deprecates the named `em_coupling` matrices in favor of a new generic structure. So correctly handling the COCOs conversions for these matrices will require deciphering the row and column URIs. I expect this will need to be a new feature in OMAS and should probably be part of a larger IMAS 4.0 upgrade

https://sharepoint.iter.org/departments/POP/CM/IMDesign/Data%20Model/sphinx/4.0/generated/ids/em_coupling.html

PR #304, opened 2024-03-29
Fixes to get machine mappings working for other experiments again

torrinba:
Source all compatible machines, protect against any that aren't, add missing dependencies, and include missing checks


orso82:
this all looks reasonable to me!

I suspect the breaking changes were introduced by some (much needed) refactoring thst @AreWeDreaming did?

If so, @AreWeDreaming can you please double check this PR?

AreWeDreaming:
I also see nothing wrong with this inherently. Do we have any known users of the machine mappings for NSTX that we can ask them to test this PR in whatever they use OMAS for?

torrinba:
The generation of k-files (which is tested in OMFIT and how I found these problems) is the only one actively used that I'm aware of.

I messaged @jmcclena about testing that for other machines (not in the OMFIT regressions) like MAST.  He might have more use cases as well

PR #308, opened 2024-06-05
fixes for MAST maching mappings.

jmcclena:
fixes for MAST maching mappings
Add MAST thomson scatter machine mapping

### DIFF
--- a/omas/omas_machine.py
+++ b/omas/omas_machine.py
@@ -6,7 +6,7 @@ import shutil
 from .omas_utils import *
 from .omas_core import ODS, dynamic_ODS, omas_environment, omas_info_node, imas_json_dir, omas_rcparams
 from .omas_physics import cocos_signals
-from omas.machine_mappings import d3d, nstx, nstxu, east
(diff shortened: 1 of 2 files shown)


torrinba (approved):
The changes look good to me. I don't have access to test with MAST data, but trust @jmcclena has already confirmed this

Unfortunately they break the unit tests unless `pyuda` is in the environment. Is that publicly available on any of our systems?

jmcclena:
@torrinba The only two systems I know that it's installed on are portal and freia. Would it make sense to wrap the import of mast machine mappings in a try-except to quickly get the regression to pass?

omas/omas_machine.py
```diff
@@ -16,6 +16,12 @@
 except:
     pass
 
+try:
+    from omas.machine_mappings import mast
+except ImportError:
+    pass
```
torrinba, line 22: Do you think it would be adding a warning message at least when MAST machine mappings are unavailable? This seems like it would be easy to miss

PR #309, opened 2024-06-11
Update the IMAS version to 3.41.0

torrinba:
This seems like a relatively small change in the parts of the schema we use most. All tests pass right out of the box.


orso82:
I assume you mean tests on your system pass? all tests in the CI seem to fail...

torrinba:
I don't see any clear reason the CI failed (I can't find a meaningful error message)

AreWeDreaming:
@orso82 Did Richard set up the CI for this or was it someone else?

smithsp:
I pushed a change here to try to get regressions to pass.

.github/workflows/regression.yml
```diff
@@ -16,7 +16,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.6', '3.7', '3.8', '3.10', '3.11']
+    # 3.9 no longer works because pint does not support it https://github.com/hgrecco/pint/issues/1974#issuecomment-2172693837
```
hassec, line 20: If I understand correctly, the error you are seeing in the CI is because it is using pint 0.23 with numpy 2.0. So wouldn't the better option be to have omas specifiy that on python version 3.9 and older you only support numpy 2.0.

Or deprecate all versions before 3.10. But supporting 3.6-3.8 and dropping 3.9 seems arbitrary given that it's an issue of dependencies being mixed incorrectly, and not an issue with python 3.9

torrinba: Thanks for calling out my laziness here. That is definitely a better solution!

hassec:
With this merged, any chance we could get a new release on pypi that includes this? 
And someonew should probably update the website to show the latest IMAS schema

torrinba:
> With this merged, any chance we could get a new release on pypi that includes this? And someonew should probably update the website to show the latest IMAS schema

Thanks for the reminder. I just updated the website.

I don't think I have permission to publish the new version to pypi, so we'll need @orso82 or @smithsp to do that

smithsp:
That was an adventure, but I have now uploaded a new pypi version.

AreWeDreaming:
Now do condaforge =D

smithsp:
Conda-forge auto-updates

hassec:
Thanks @smithsp :bow:

PR #316, opened 2024-07-31
Remove pyuda warning

### LINKED ISSUE #315: Avoid MASTU related warning on import
I think it's a bit unfortunate that since `0.94` a simple
```
import omas
```

will print:
```
Could not import mast machine mappings. Check that pyuda is installed
```

That seems like a warning that shouldn't be presented to every user, but only those that actually want to use those machine mappings.

Given that `omas/__init__.py` has a `from .omas_core import *`, disentangling these imports might be tricky. But then I propose to at least remove the noisy print.

hassec:
closes #315

### DIFF
--- a/omas/omas_machine.py
+++ b/omas/omas_machine.py
@@ -19,7 +19,6 @@ except:
 try:
     from omas.machine_mappings import mast
 except ImportError:
-    print('Could not import mast machine mappings. Check that pyuda is installed')
     pass
 
 


hassec:
@orso82 any chance we can get a pypi release that includes this, so I can push and update to avoid users having to worry about this warning?

orso82:
v0.94.1 should be on pypi now

hassec:
Awesome, thanks for getting this done so quickly @orso82 🙏

PR #320, opened 2024-10-25
Robust plotting of profiles and their data from ZIPFIT and OMFIT_PROFS MDSplus trees

smithsp:
The removal of whitespace can be ignored with adding `?w=1` to the files tab.
The bugs were discovered while trying to plot profiles for the SC24 Demo (of CAKE and IRI).

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -47,14 +47,14 @@ jobs:
     # but before pygacode 1.0 it did not specify its build dependencies correctly
     # and pygacode requires python >= 3.8,
     # so for 3.7 we need to build without build isolation and manually install numpy
-    - name: Install OMAS (Py 3.7)
-      if: ${{matrix.python-version == '3.7' || matrix.python-version == '3.8' || matrix.python_version == '3.9'}}
+    - name: Install OMAS (Py 3.7, 3.8, 3.9)
+      if: ${{matrix.python-version == '3.7' || matrix.python-version == '3.8' || matrix.python-version == '3.9'}}
       run: |
         python3 -m pip install wheel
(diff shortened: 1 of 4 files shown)


.github/workflows/regression.yml
```diff
@@ -47,14 +47,14 @@ jobs:
     # but before pygacode 1.0 it did not specify its build dependencies correctly
     # and pygacode requires python >= 3.8,
     # so for 3.7 we need to build without build isolation and manually install numpy
-    - name: Install OMAS (Py 3.7)
-      if: ${{matrix.python-version == '3.7'}}
+    - name: Install OMAS (Py 3.7, 3.8, 3.9)
+      if: ${{matrix.python-version == '3.7' || matrix.python-version == '3.8' || matrix.python-version == '3.9'}}
```
hassec, line 51: Why do 3.8 and 3.9 now also need to be built w/o build isolation?

smithsp: I don't know exactly. I was getting an error about a compiler not being found (you can check the logs for the previous commits), and hoped that this fix would resolve the problem, and it did.  I am not clear about what using this option does or what the implications are.  If you have other suggestions, I am open to them.  I would like to merge this very soon, as I have a live demo at the Supercomputing conference on Nov 18 that depends on these fixes.  Of course I could work off of a branch, but that is not as preferable as leaving this pull request dangling.

hassec: It looks like it's again a build problem, with pygacode. But I don't have access to that package, so these are hard to debug.

smithsp: Thanks @hassec .

PR #321, opened 2024-11-12
Add toray bhalf

AreWeDreaming:
Small addition needed for https://github.com/gafusion/OMFIT-source/pull/7203 .
It adds the `bhalf` factor that Toray uses in order to pretend to be a beam tracing code. Before it was hard-coded to `1.7 deg.` in OMFIT but now it is loaded from MDS+ and passed through properly.

We are still not loading the Gaussian optics parameters from MDS+ and Michael Ross promised me that we'll look into this in December.


jmcclena (approved):
works for me

PR #324, opened 2024-11-14
test_machine_mapping_functions under each machine mappings

orso82:
I think having the tests at the bottom of each machine mapping file makes more sense.
While `test_machine_mapping_functions` worked for DIII-D it was broken for other devices.

### DIFF
--- a/omas/machine_mappings/d3d.py
+++ b/omas/machine_mappings/d3d.py
@@ -1523,3 +1523,5 @@ def core_profiles_global_quantities_data(ods, pulse):
 
 
 # ================================
+if __name__ == '__main__':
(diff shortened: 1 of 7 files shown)


AreWeDreaming:
Do we really want any test function in the actual file and not in `omas/tests`. I can move the tests there if @orso82 agrees.

orso82:
That will not work, since you cannot access MDS+ data for those machines. If `test_machine_mapping_functions()` is confusing, it could be renamed as `run_machine_mapping_functions()`.

AreWeDreaming (approved):
I am approving this, but I also opened https://github.com/gafusion/omas/issues/325 so we can find a better way to run these tests.

PR #327, opened 2024-11-22
IMAS 4.0.0

orso82:
I have pushed a branch with v4.0.0 of the data dictionary. However transitioning to v4 requires some thought. For existing software, the use of `COCOS 17` instead of `COCOS 11` is particularly problematic, and the change of `_tor` to `_phi` is also very annoying.

Beware that if you try to use this branch as it is, the `cocos.py` file is not updated for v4.0.0 thanks to the `_tor` to `_phi` change. Also all machine mappings will need to be reviewed and updated.

One option going forward may be transition to `v4` but add some automated translation between v3 and v4 within OMAS itself. OMAS automatic COCOS translation `with omas_environment(ods, cocosio=???):` allow us to gracefully handle the COCOS 11 to 17 change. To avoid creeping of bugs we could change OMAS to raise an error if a quantity that is affected by the COCOS change gets accessed ouside of a `with omas_environment`. Then, all we need to do then is to handle `_tor` to `_phi` translations, which would be doable.

Anyways, before rushing to make in any such update, I'd like to wait and see how quickly the IMAS ecosystem adapts to the v4 changes.


kripnerl:
ping to `github-actions`

PR #330, opened 2025-01-14
make heavy dependency optional

### LINKED ISSUE #261: OMAS enhancement proposals
Hi, 

since I have been working with OMAS package lately, I have found some things that may be pretty handy to implement.

Feel free to check the Streamlit app, even though it is trivial. I find it pretty handy.  

# Support "+" operator 

This one is simple: 
```
ods1 = omas.ODS()
ods1.sample_magnetics()

ods2 = omas.ODS()
ods2.sample_pf_active()

ods3 = ods1 + ods2
```

An implementation may be pretty straightforward: 

```python
def plus(ods1, ods2):
    ods_res  = omas.ODS()
    ods_res.update(ods1)
    ods_res.update(ods2)
    return ods_res
```
not sure how memory heavy this is... 

# Support ByteIO in "load" function. 

This one is quite important with the increase in the number of web services. I have created a simple Streamlit OMAS viewer: 

```python
import os

import omas
from omas import ODS
from omas.omas_utils import json_dumper
from tempfile import NamedTemporaryFile

import streamlit as st
import json

import matplotlib.pyplot as plt


def ods2json(ods):
    kw = dict()
    kw.setdefault('indent', 0)
    kw.setdefault('separators', (',', ': '))
    kw.setdefault('sort_keys', True)

    json_string = json.dumps(ods, default=lambda x: json_dumper(x, None), **kw)

    return json_string


def load_uploaded_omas_file(file):
    _, suffix = os.path.splitext(file.name)

    with NamedTemporaryFile(suffix=suffix, delete=False) as temp:
        temp.write(file.getvalue())
        temp_file_path = temp.name

    ods = ODS().load(temp_file_path)
    os.remove(temp_file_path)
    return ods


ods_file = st.file_uploader("Choose OMAS IDS")

if ods_file is not None:
    plot_tab, tree_tab = st.tabs(["Plot", "Tree"])

    # ods_obj = ODS().load("./machine@2022.05v2.h5")
    ods_obj = load_uploaded_omas_file(ods_file)

    with plot_tab:
        fig = plt.figure()
        ods_obj.plot_overlay(wall=True, pf_active=True, pf_passive=True)
        st.pyplot(fig)

    with tree_tab:
        ods_json = ods2json(ods_obj)
        st.json(ods_json, expanded=False)
```

image

with installed `Streamlit`, you can run it as `streamlit run app.py`. 

However, you may notice `load_uploaded_omas_file` function here. I need to create a temporary file, save and reload the downloaded IDS. I can imagine a lot of similar applications when working with IDS/OMAS data. 

# `ods2json` function. 
As mentioned above, working with JSON "in" memory" is handy. Esp. when JSON saving function supports only the filename and no file object.  

# Support `pathlib` 

Fallowing example 
```python
import pathlib
import omas

path = pathlib.Path("my_omas.h5")
ods = omas.ODS()
ods.load(path)
```
should be equivalent to
```python
# ...

ods.load(str(path))
# or
ods.load("my_omas.h5")
```

### LINKED ISSUE #280: Make some serializers optionals
# Reasoning
OMAS should be (IMHO) a lightweight tool for; however, some of its dependencies make it bulky and have complicated dependencies. 

# Problem

I speak specifically about `mojo-core`.  We have already encountered a second issue with that library. First was an obsolete dependency on urllib (which is fixed in the newer version) and second is now Windows poetry installations: 

```
Installing botocore (1.31.33): Failed

  FileNotFoundError

  [Errno 2] No such file or directory: 'C:\\Users\\jh\\AppData\\Local\\Packages\\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\\LocalCache\\Local\\pypoetry\\Cache\\virtualenvs\\cug-vessel-model-WXhTS8yu-py3.11\\Lib\\site-packages\\botocore\\data\\kinesis-video-archived-media\\2017-09-30\\endpoint-rule-set-1.json.gz'

  at C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.11_3.11.1776.0_x64__qbz5n2kfra8p0\Lib\pathlib.py:1044 in open
      1040│         the built-in open() function does.
      1041│         """
      1042│         if "b" not in mode:
      1043│             encoding = io.text_encoding(encoding)
    → 1044│         return io.open(self, mode, buffering, encoding, errors, newline)
      1045│
      1046│     def read_bytes(self):
      1047│         """
      1048│         Open the file in bytes mode, read it, and close the file.
```

# Proposed solution

Add specialized backend as optional dependencies in `setup.py`:

```python
...
install_requires = [
    'numpy>=1.16.1',
    'uncertainties',
    'pint',
    'netCDF4',
    'matplotlib',  # In principle, plotting could be also optional since some deployments do not require any plotting.
    'scipy',
    'h5py',
    'dnspython', # Not sure about this one - it is not used in the code, or I missed it. 
    'xarray',
    'setuptools>=41.2',  
    'Cython',
]

extras_require = {
    'machine': [
        'omfit_classes',
        'pexpect',
        'fortranformat',
        'pygacode',
    ],
    'hdc': ['pyhdc'],
    'imas': ['imas'],
    'uda': ['pyuda'],
    'mongo': ['pymongo'],
    'aws': ['boto3'],
    
    'build_structures': ['bs4', 'xmltodict', 'tqdm'], # xmltodict is used only in `omas_structures` and `tqdm` only when fetching data and the code more over is checking on the existence of the `tqdm`.  
    'build_documentation': ['Sphinx', 'sphinx-bootstrap-theme', 'sphinx-gallery', 'Pillow'],
}
...
```

kripnerl:
Closes #280

- [ ] Divide `S3` and `mongo` dependencies into eventual groups. 
- [ ] Remove obsolete dependencies. 
- [ ] Add installation instructions.

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -42,7 +42,7 @@ jobs:
     - name: Install dependencies
       run: |
         python3 -m pip install -r requirements.txt
-        python3 -m pip install boto3 pymongo
+        python3 -m pip install boto3 pymongo xmltodict
 
     # numpy is a build time dependency of pygacode
     # but before pygacode 1.0 it did not specify its build dependencies correctly
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -42,6 +42,7 @@ jobs:
     - name: Install dependencies
       run: |
         python3 -m pip install -r requirements.txt
+        python3 -m pip install boto3 pymongo
 
     # numpy is a build time dependency of pygacode
     # but before pygacode 1.0 it did not specify its build dependencies correctly
--- a/.github/workflows/regression_no_munittest.yml
+++ b/.github/workflows/regression_no_munittest.yml
@@ -29,7 +29,7 @@ jobs:
 
     - name: Install OMAS
       run: |
-        python3 -m pip install .[machine]
+        python3 -m pip install .[machine, s3, mongo]
 
     - name: Move to root and try importing OMAS
       run: |
--- a/.github/workflows/regression_no_munittest.yml
+++ b/.github/workflows/regression_no_munittest.yml
@@ -26,10 +26,11 @@ jobs:
     - name: Install dependencies
       run: |
         python3 -m pip install -r requirements.txt
+        python3 -m pip install boto3 pymongo
 
     - name: Install OMAS
       run: |
-        python3 -m pip install .[machine, s3, mongo]
+        python3 -m pip install .[machine]
 
     - name: Move to root and try importing OMAS
       run: |
--- a/.github/workflows/regression_no_munittest.yml
+++ b/.github/workflows/regression_no_munittest.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Install dependencies
       run: |
         python3 -m pip install -r requirements.txt
-        python3 -m pip install boto3 pymongo
+        python3 -m pip install boto3 pymongo xmltodict
 
     - name: Install OMAS
       run: |
--- a/imports_check.py
+++ b/imports_check.py
@@ -10,7 +10,7 @@ import warnings
 warnings.simplefilter('ignore', category=DeprecationWarning)
 warnings.simplefilter('ignore', category=FutureWarning)
 
-mapper = {'dnspython': None, 'sphinx-bootstrap-theme': None, 'sphinx-gallery': None, 'Sphinx': 'sphinx', 'Pillow': 'PIL'}
+mapper = {'sphinx-bootstrap-theme': None, 'sphinx-gallery': None, 'Sphinx': 'sphinx', 'Pillow': 'PIL'}
 
 filename = os.path.split(os.path.abspath(__file__))[0] + os.sep + 'requirements.txt'
 with open(filename, 'r') as f:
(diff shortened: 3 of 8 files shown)


kripnerl:
By the way, I did quite an ugly hack with tests to run. I'm not sure whether to leave it like this or find some more elegant solution. 

If there are `pytest` I would add some flags to run some particular tests only if the selected package is installed. To be sure, that the library is not imported where it shouldn't be. 

Nevertheless, I would appreciate this request to be merged as soon as possible since the dependency hell, which is (most likely) `botocore` causing is blocking me quite a lot - poetry lock file is being calculated for over several minutes. Otherwise it takes only few seconds.

orso82:
I don't think s3 and mongo are used much, if at all. I see regression tests passing, and I am supportive to merge.
@AreWeDreaming @smithsp @torrinba @hassec thoughts?

AreWeDreaming:
I certainly don't use either of those with OMAS directly, but then I don't do a lot of database work overall.

hassec:
Thanks for the ping @orso82, no objections from my side 👍

smithsp:
Good by me.  Good to know.

PR #334, opened 2025-02-24
Correct and verified conversion to ITER launch coordinate system

AreWeDreaming:
This changes the DIII-D mapping for the ECH to use the correct conversion from DIII-D to ITER angles. The calculation behind this was graciously provided by E. Poli. I have checked the initial wave vector in Torbeam against the wave vector in TORAY. 
TORAY uses:
kx = sin a cos b
ky = sin a cos b
kz = cos a
with a = t + pi/2 and b = p + pi one gets
kx = -cos(t) cos(p)
ky = -cos(t) sing(p)
kz = sin(t)
Which is the same as in Torbeam with the exception of kz which has the opposite sign. This could mean that the coordinate system in TORAY points in the opposite direction in the z direction. In practice the two codes agree perfectly now:
image
[notes_angle_convention.pdf](https://github.com/user-attachments/files/18947816/notes_angle_convention.pdf)

### DIFF
--- a/omas/machine_mappings/d3d.py
+++ b/omas/machine_mappings/d3d.py
@@ -556,8 +556,11 @@ def ec_launcher_active_hardware(ods, pulse):
         else:
             beam['time'] = np.atleast_1d(gyrotrons[f'TIME_AZIANG_{system_no}']) / 1.0e3
         ntime = len(beam['time'])
-        beam['steering_angle_tor'] = np.atleast_1d(np.deg2rad((gyrotrons[f'AZIANG_{system_no}'] - 180.0)))
-        beam['steering_angle_pol'] = np.atleast_1d(np.deg2rad((gyrotrons[f'POLANG_{system_no}'] - 90.0)))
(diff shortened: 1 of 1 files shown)


AreWeDreaming:
Note that this MUST not be merged into OMFIT before we thoroughly go through to code and check where the OMAS launchers are used.

AreWeDreaming:
@jmcclena if I look for `ec_launchers` in OMFIT source there aren't too many hits outside of the TORAY module with the exception of `CHEF` and `STEP`. Will you look into those or do you need my help?

smithsp:
This is passing regressions.  Is it ready to merge?

AreWeDreaming:
Recently the ECH group added some Gaussian beam parameters (phase front curvature radius and spot size) to MDSplus. I am now attempting to fetch these quantities from MDSplus and now only use the hardcoded values as a backup. Effects on results is negligible with the new Gaussian beam parameters for current gyrotrons.

AreWeDreaming:
The agreement is even better now:
image

PR #336, opened 2025-03-11
Bug fix for DIII-D profiles mapping

torrinba:
The problem was uncovered running the OMFIT `iri_eq_plotter` which was tested and merged last week so I'm not sure how this crept in.


smithsp (approved):
Sorry for breaking anything.

smithsp:
@torrinba Are you saying that testing the `iri_eq_plotter` on OMFIT's unstable branch was failing?  I think that this change will work for the latest changes I made to that script and here in omas, but just want to understand how you came across this problem.

torrinba:
> @torrinba Are you saying that testing the `iri_eq_plotter` on OMFIT's unstable branch was failing? I think that this change will work for the latest changes I made to that script and here in omas, but just want to understand how you came across this problem.

No that's a good point. I was actually still using an older version of `iri_eq_plotter` that I had tested before the changes you made ahead of the merge. I can see that you modified the script to avoid this error as well so it is working on unstable. This change won't break that, but it does mean fetching profiles can use arguments that are more like the other mappings.

PR #338, opened 2025-03-28
updates to D3D mappings

orso82:


torrinba (approved):
I'm not familiar with the fetching NBI data through OMAS, but the `function` definitions for pf coils are appropriate

PR #342, opened 2025-05-07
Attempt to use Ubuntu 24.04 for tests

torrinba:
This seemed simple enough to try out.

Issue #341, opened 2025-05-07
Regression Test Runner Needs To Be Updated


torrinba:
Tests aren't being run by the merge pipeline any longer because the Ubuntu version appears to be too old. Hopefully this is a trivial fix for whoever created the pipeline.

![Image](https://github.com/user-attachments/assets/052cc51e-a168-4af4-bc63-aeee0bffe470)

torrinba:
#342 is an attempt at fixing this, but doesn't work for all Python versions. There seems to be an assumption that the minor Python version is 0, but that isn't supported for 3.6 and 3.7. Specifying the minor version directly causes other problems though.

There is also a note in `regression.yml` about pygacode not supporting Python 3.12 because numpy.distutils was removed. I'm not sure how pygacode is used here, but if that hasn't been addressed it seems like a work around will be required

PR #339, opened 2025-04-23
cannot assume homogeneous time for D3D magnetics_probes_data

orso82:

PR #343, opened 2025-05-07
Update regressions to use Ubuntu 22.04

torrinba:
Addresses #341.

This requires removing the Python 3.6 test since that version is no longer supported.

A larger jump to Ubuntu 24.04 is also possible (attempted in #342), but requires Python 3.7 to be retired as well. Since OMFIT is still on 3.7, this doesn't seem like a good idea yet.


torrinba:
Closing in favor of #339

PR #345, opened 2025-05-12
Reuse MDSplus connections

AreWeDreaming:
Before this PR OMAS would open a separate connection for every `(server, tree, shot)` tuple. This makes it impossible to retrieve more than `1000` unique tuples in the same session for DIII-D because Atlas enforces a 1000 connection limit.
With this change Omas cashes only a single connection and `(tree,.shot)`. So as long as you operate within the same `(tree,shot)` tuple there is no functional difference. If you move between many `(tree,shot)` tuples then this should be faster as the connection is reused. If, however, you just constantly flip-flop between two `(tree1, shot1)`, `(tree2, shot2)` tuples then it will be slower since we are constantly opening and closing trees. 
I propose we make this change even though because this can have negative side effects because it reduces footprint on Atlas and it follows best practices more than the original version.

There are a couple upgrades that can be made. My hesitation is though that the main reason for these is upgrades is to compensate for bad user patterns. In an ideal user pattern the user gets all data from one `(tree,shot)` tuple and then they move to the next. If someone has a good use-case that would require the flip-flopping between `(tree,shot)` tuples then I can see if we can handle those specifically.

### DIFF
--- a/omas/utilities/omas_mds.py
+++ b/omas/utilities/omas_mds.py
@@ -8,6 +8,19 @@ __all__ = [
     'get_pulse_id'
 ]
 
+last_tree = None
+current_conn = None
+
+def manage_open_tree(conn, treename, pulse):
+    global last_tree
+    if last_tree == (treename, pulse):
+        return
+    elif last_tree is not None:
+        conn.closeTree(*last_tree)
+    conn.openTree(treename, pulse)
+    last_tree = (treename, pulse)
+
+
 def get_pulse_id(pulse, run_id=None):
     """
     Converts the pulse number into a MDSplus run_id
@@ -67,9 +80,6 @@ def tunnel_mds(server, treename):
         return tunneled_server
 
     return server.format(**os.environ)
-
-
-
 
 
 class mdsvalue(dict):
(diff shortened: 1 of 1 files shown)


AreWeDreaming:
Ok so after some more conversation with Sean he advised me to not manually manage the trees and it makes a huge difference. Now after testing this for the CAKE database letting MDSplus track the open trees internally is 20 times faster. So this should be the best of both worlds.

omas/utilities/omas_mds.py
```diff
@@ -146,22 +146,18 @@ def mdsk(value):
...
                 out_results = None
-
-                # try connecting and re-try on fail
+                global current_conn
                 for fallback in [0, 1]:
-                    if (self.server, self.treename, self.pulse) not in _mds_connection_cache:
-                        conn = MDSplus.Connection(self.server)
-                        if self.treename is not None:
-                            conn.openTree(self.treename, self.pulse)
-                        _mds_connection_cache[(self.server, self.treename, self.pulse)] = conn
```
smithsp, line 156: Couldn't the overall changes of this PR be accomplished by just changing the key for the cache to `self.server` (and moving some of the logic here)?  I don't like the `current_conn` global; what if a person wants to connect to more than one server simultaneously?

smithsp (changes requested):

AreWeDreaming:
Seems to work just as well with the change.

AreWeDreaming:
Do we know what's wrong with `test_omas_mongo_example` for python 3.7?

torrinba:
> Do we know what's wrong with `test_omas_mongo_example` for python 3.7?

It's temperamental...

smithsp (approved):

smithsp:
This is passing.  Do we need a simultaneous OMFIT branch with these changes?

AreWeDreaming:
No, this is purely an under-the-hood change.

AreWeDreaming:
Nvm found the bug, no idea what I was thinking there...

smithsp:
The failing example should go into a test.

PR #348, opened 2025-05-28
Reattempt to implement the new MDSplus connection management

AreWeDreaming:
The problem with the original attempt was that when trying to get signals from tree A and tree B one has to explicitly reopen tree B after getting some signals from tree A even if tree B was previously opened. Since ECH launchers fetches first from RF then from EFIT then again from RF it breaks while other tests work just fine.


smithsp:
@AreWeDreaming Can you be more explicit about the command to run?

torrinba (approved):
This works for all of the OMAS mappings used in OMFIT that I'm aware of https://github.com/gafusion/OMFIT-source/pull/7385

That probably doesn't includes everything, but it's enough that I would be comfortable with merging this again.

I also tried running
```
ods = load_omas_machine('d3d', 200000)
```
to test every mapping, but it gets hung up because of missing equilibrium data and isn't really a good test right now. I'd like to get that working and add it to the OMFIT regression tests, but it will take some time to make things robust enough.

The `test_machine_mapping_functions` routine that is run when `d3d.py` is executed requires the mapping functions being tested to be passed as arguments. Currently two are hardcoded in `d3d.py`, but the default when `__all__` is specified is to not call any functions. So unless there's a way to specify that everything defined should be run, it doesn't seem very useful.

torrinba:
I'm fine with merging now, but like the previous attempt where something was overlooked, the full set of DIII-D machine mappings hasn't been tested with this change. So if @AreWeDreaming or @orso82 wants to try more of them first it could uncover something else that wasn't expected. Alternatively it could wait for the new OMAS test I want to setup in OMFIT that fetches all of the DIII-D mappings available which I hope to get to in the next week or two.

AreWeDreaming:
@torrinba I would hold off on this because the toksearch integration will introduce some drastic changes on how mdsplus connections are handled in OMAS.

AreWeDreaming:
I ran into this issue again when trying to count CAKE equilibria (too many MDSplus connection). I checked and while my toksearch changes touch many things, they do not touch how trees are cached so we need to get these changes in after all. Thanks for @torrinba for keeping this ball rolling. I am testing this now for fetching the time base of the entire CAKE database. That should be a pretty good robustness test.

AreWeDreaming:
The CAKE fetching went well (and fast). I also tested this in OMAS viewer where it works fine as well. I think this is ready to go.

PR #353, opened 2025-06-24
Add option for strict grid limit enforcement in core_profiles

AreWeDreaming:
This PR adds strict grid limits for `core_profiles` that apply when mapping from the `OMFIT_PROFS` tree. At the moment the default is to enforce these limits but more discussion is needed to decide if this is the best option.

In OMFIT `rho`, an interpolation between `rho_tor` in the core and `rho_pol` in the SOL, is used pretty consistently in lieu of a proper `rho_tor`. Hence, using the same convention in OMAS makes sense to me.

On the other hand this is inconsistent with the IMAS scheme which is certainly not ideal either.

Personally, I don't like the idea of hiding away parts of the profile just because the IMAS schema makes it inconvenient to store it. If we use an optional keyword that defaults to limiting profiles to rho_tor < 1.0 then no one will ever find the option. Even auditing all 200 references to `core_profiles` in OMFIT might be insufficient because in many cases the ODS is loaded manually by the user. If they use the mappings to create the ODS themselves it is completely out of our hands. Maybe some documentation could help but I am not convinced.

Below profile plots with the option turned on/off:
image

image

### DIFF
--- a/omas/machine_mappings/d3d.json
+++ b/omas/machine_mappings/d3d.json
@@ -15,7 +15,8 @@
   "fast_ece": false,
   "get_all": true,
   "nref": 0,
-  "revision": "BLESSED"
+  "revision": "BLESSED",
+  "core_profiles_strict_grid": true
  },
  "bolometer.channel.:": {
   "PYTHON": "bolometer_hardware(ods, {pulse})"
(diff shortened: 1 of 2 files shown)


torrinba (approved):
This didn't cause any OMFIT regression tests to fail for the modules mentioned or any that I'm aware would use OMAS profiles https://github.com/gafusion/OMFIT-source/pull/7384. I'll leave the OMFIT PR open in case you think of any others that needs to be tested, but I feel pretty comfortable that changing the default behavior won't be catastrophic for OMFIT (I'm sure there are some special cases that don't have regression tests, but I also don't think that should be a top concern either)

AreWeDreaming:
I would be pretty surprised if any of the regression0 tests  in OMFIT actually touch the `OMFIT_PROFS` tree.

AreWeDreaming:
@quinntpratt Do you need OMAS mapping that extend beyond the LCFS for any of your work?

AreWeDreaming:
Since there has been no comment I am closing this in favor of moving ahead with #354

PR #355, opened 2025-06-26
Add fuse data export script

AreWeDreaming:
This moves some OMAS related python source code from FUSE to OMAS. Since this is a quite comprehensive data export I thought it was more useful in OMAS than tugged away in FUSE.
EDIT: Changed base to `d3d_map_updates` since they are reasonably related.

### DIFF
--- a/omas/machine_mappings/d3d.py
+++ b/omas/machine_mappings/d3d.py
@@ -595,7 +595,7 @@ def ec_launcher_active_hardware(ods, pulse):
 
         beam['frequency.time'] = np.atleast_1d(0)
         if isinstance(systems[f'FREQUENCY_{system_no}'], Exception):
-            beam['frequency.data'] = beam['frequency.time'] * 0.0 + 110e9 # old shots did not record the frequency, since they were all at 110 GHz
+            beam['frequency.data'] = np.ones(beam['frequency.time'].shape) * 110e9 # old shots did not record the frequency, since they were all at 110 GHz
         else:
             beam['frequency.data'] = np.atleast_1d(systems[f'FREQUENCY_{system_no}'])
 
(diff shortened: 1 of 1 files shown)


AreWeDreaming:
To test:
```
python omas/examples/fuse_data_export.py omas_test.h5 193843 EFIT02 ZIPFIT01
python omas/examples/fuse_data_export.py omas_test.h5 193843 EFIT OMFIT_PROFS --EFIT_RUN_ID 08 --PROFILES_RUN_ID 007
```

orso82:
@AreWeDreaming

`fuse_data_export` --> `d3d_data_export`

or maybe

`python omas/examples/fuse_data_export.py omas_test.h5 D3D 193843 EFIT OMFIT_PROFS --EFIT_RUN_ID 08 --PROFILES_RUN_ID 007`

since I can see we will eventually want to use this for different devices.

Is your idea that FUSE should be using this from now on? If that's the case, I want to converge on a API that is generic and stable enough that will handle different versions of FUSE calling this script. How do we handle new arguments (easy, with defaults) and deprecated arguments?

AreWeDreaming:
I am still working on the accompanying FUSE PR that will then use this script.
As for the arguments: We can simplify a little bit by having OMAS (with some help from OMFIT_rdb) try to find the CAKE run ids using a priority list for the various CAKE flavors like this ['CAKE_FDP', 'CAKE02', 'CAKE01']. That should be easy enough to set up and streamlines the process. On the FUSE side it would then be just a switch with `CAKE=false` as the default. 

As for other arguments I am not sure what direction you are thinking and how to handle future upgrades depends a lot on the upgrades. For now I think this is generic enough and I'm happy to generalize it when modifications in FUSE warrant it.

orso82:
My only concern is to have FUSE rely on some interface that may change in the future. Of course we can make the latest FUSE release to work with the latest API, but that may leave older FUSE versions broken... I am not saying no to this, just that the API for the `fuse_data_export.py` script should be well thought out and somewhat future proof.

AreWeDreaming:
I still have the hope that this will ultimately be a stop-gap measure and we will get revised mappings that will have a robust and stable API

omas/examples/fuse_data_export.py
```diff
@@ -0,0 +1,97 @@
...
+    # printe("- Fetching nbi data")
+    # d3d.nbi_active_hardware(ods, shot)
+
+    printe("- Fetching core_profiles data")
+    d3d.core_profiles_profile_1d(ods, shot, PROFILES_tree=PROFILES_TREE, 
+                                 PROFILES_run_id=PROFILES_RUN_ID)
+
+    printe("- Fetching wall data")
+    d3d.wall(ods, shot)
+
```
torrinba, line 34:
```suggestion

        printe("- Fetching magnetic hardware data")
        d3d.magnetics_hardware(ods, $shot)

```
Include missing lines from the FUSE master branch

PR #354, opened 2025-06-25
D3D mapping updates (includes #353)

orso82:

PR #356, opened 2025-06-27
update d3d machine mappings

jmcclena:
remove zeroing of pf_active
add rotation to zipfits

### DIFF
--- a/omas/machine_mappings/d3d.py
+++ b/omas/machine_mappings/d3d.py
@@ -1332,9 +1332,6 @@ def magnetics_floops_data(ods, pulse, nref=0):
             data_norm=1.0,
         )
 
-    # Set reference flux loop to zero before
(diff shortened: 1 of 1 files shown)


jmcclena:
I also tested the cocos, and found that zip rotation is positive for
```
            shottime[(150219, 3200)] = (+1, +1)  # +Bt +Ip
            shottime[(133221, 1151)] = (-1, +1)  # -Bt +Ip
            shottime[(153298, 4400)] = (-1, -1)  # -Bt -Ip
```

shot 153298 is a QH module with injection only from 30L and 30R (traditionally co-beams).

orso82:
Thank you @jmcclena !
Please note that I have changed the base of this PR to the `d3d_map_updates` branch for which we have #354 open

PR #359, opened 2025-07-18
Improvements to fetch all mapped data

torrinba:
These changes were required to run
`load_omas_machine('d3d', 200000)`
without any errors. Some of the fields aren't filled, but these are now skipped or zeroed without any hard failures.


smithsp (approved):
One minor suggestion (if I read it right); otherwise good to merge.

PR #361, opened 2025-07-22
Add files that were missed in the original attempt

AreWeDreaming:
I missed several files when I initially made this PR. They are now included. I have tested one example but a more thorough test might be good this time around.


AreWeDreaming:
I think the `summary.global_quantities` is more appropriate

torrinba:
> I think `summary.global_quantities.v_loop` is more appropriate but it implies that this is preprocessed data similarly to `core_profiles`. I am really surprised that I cannot find a purely diagnostic IDS for this quantity.

Was this comment intended for https://github.com/gafusion/omas/issues/362?

torrinba:
It looks like the mappings run now, but I check the values and see some clear problems with the weights. In the fitwieght.dat file the are all ones and zeros so something is clearly going wrong.
weight
The compensation values should probably be checked carefully as well, but won't be quite as simple

torrinba:
The `core_profiles` machine mappings are no longer crashing, but that doesn't mean all of these changes are working either. Since the mapping gets the diagnostic weights from the EFIT tree (not the fitweight file) those need further testing. When I try running 
```
fitweight = D3DFitweight(support_filenames('d3d', 'fitweight', 200000))
```
the result is an empty dictionary. But 
```
fitweight = OMFITd3dfitweight(support_filenames('d3d', 'fitweight', 200000))
```
works fine, so there appears to be a problem with the translation. The same issue exists for the compensations so the those machine mappings should be failing now.

omas/machine_mappings/_common.py
```diff
@@ -9,6 +9,108 @@
...
+    def __init__(self, filename):
+        r"""
+        OMFIT class to parse DIII-D device files
+
+        :param filename: filename
+
+        :param \**kw: arguments passed to __init__ of OMFITascii
+        """
+        self.filename = filename
+        self.load()
```
AreWeDreaming, line 26: You shouldn't do this because the following syntax is used to load these files:
```
def get_support_file(object_type, filename):
    """
    Cached loading of support files

    :param object_type: Typically a OMFIT class

    :param filename: filename of the support file to load
    """
    if filename not in __support_files_cache__:
        __support_files_cache__[filename] = object_type(filename)
        __support_files_cache__[filename].load()
    return __support_files_cache__[filename]
```

```suggestion
```

omas/machine_mappings/_common.py
```diff
@@ -9,6 +9,108 @@
...
+    def __init__(self, filename):
+        r"""
+        OMFIT class to parse DIII-D MHD device files
+
+        :param filename: filename
+
+        :param \**kw: arguments passed to __init__ of OMFITascii
+        """
+        self.filename=filename
+        self.load()
```
AreWeDreaming, line 90:
```suggestion
        self.load()
```
Same as above

omas/machine_mappings/d3d.json
```diff
@@ -369,6 +369,21 @@
...
+ },
+ "equilibrium.time_slice.:.constraints.bpol_probe.:.weight": {
+  "PYTHON": "magnetics_weights(ods, {pulse}, {time_index!r})"
+ },
+ "equilibrium.time_slice.:.constraints.flux_loop.:": {
+  "PYTHON": "magnetics_weights(ods, {pulse}, {time_index!r})"
+ },
+ "equilibrium.time_slice.:.constraints.flux_loop.:.weight": {
+  "PYTHON": "magnetics_weights(ods, {pulse}, {time_index!r})"
+ },
```
torrinba, line 386:
```suggestion
```
This overrides the mappings for these same variables in `_efit.json` making them incompatible. Since it also requires the `time_index` to be specified it isn't as useful in most fetches (like `machine_to_omas`).

torrinba (approved):
Runs properly and everything looks like it has been fixed to me

PR #368, opened 2025-08-05
Ensure that DIII-D flux loop data is consistent with IMAS

torrinba:
DIII-D measures and stores all other flux loops differentially relative to PSF1A and has since at least 1988. Unfortunately the only differential flux loops supported in IMAS are diamagnetic so the poloidal loops must be converted to total flux in order to be valid. This is now the default behavior, but an option is available to skip the conversion if IMAS adherence isn't a concern. The option to zero the reference flux loop has been completely removed since it throws out important information. Support for differential poloidal flux loops, which should only require minor changes in the data schema, has been requested through GitHub https://github.com/iterorganization/IMAS-Data-Dictionary/issues/128 . Even if IMAS adds support for this, it probably won't be in a version supported by OMAS in the foreseeable future (since there are no plans to move to v4 where most schema changes are being made).

Also moved the flux loop and probe compensations after the uncertainties are fetched to avoid attempting their application for channels that are disabled. This hopefully eliminates the need for the magnetics changes in #366 (I would be happy to test if @orso82 can provide a shot number where problems with the compensations are seen)


bclyons12 (commented):
I have some concern about generally using COCOS 7, as it might not be true for every diagnostic. Just be careful here, and compare the ODS you get with COCOS 1 vs 7 to make sure that only the values you wanted changed are affected.

torrinba:
> I have some concern about generally using COCOS 7, as it might not be true for every diagnostic. Just be careful here, and compare the ODS you get with COCOS 1 vs 7 to make sure that only the values you wanted changed are affected.

Confirmed (this is for all mappings pulled by FUSE):
```
h5diff cocos_7.h5 cocos_1.h5 
dataset: </magnetics/flux_loop/0/flux/data> and </magnetics/flux_loop/0/flux/data>
30639 differences found
dataset: </magnetics/flux_loop/1/flux/data> and </magnetics/flux_loop/1/flux/data>
30703 differences found
dataset: </magnetics/flux_loop/10/flux/data> and </magnetics/flux_loop/10/flux/data>
30715 differences found
dataset: </magnetics/flux_loop/11/flux/data> and </magnetics/flux_loop/11/flux/data>
30718 differences found
...
dataset: </magnetics/flux_loop/9/flux/data> and </magnetics/flux_loop/9/flux/data>
30712 differences found
```

torrinba:
I would like to merge since the corresponding changes anticipating this format have been merged in FUSE

torrinba:
> I presume the new mappings have been tested against conventional methods for getting this data, e.g. EFIT inputs?

That is ongoing, but involves time averaging the data so I haven't done a quantitative comparison. Taking the nearest time point gives similar values once the COCOs factors are accounted for

jmcclena (approved):
Looks fine to me. The EGGO CAKENN from dd routine might need to be updated once this is live.

orso82:
@jmcclena note that this is already live when we pull data from DIII-D via FUSE

PR #371, opened 2025-08-21
Move contourpy dependency to the function for OMFIT compatibility

AreWeDreaming:
Simple fix to address missing dependency in OMFIT


torrinba:
It's worth noting that this dependency will need to be added to OMFIT when the toksearch branch is merged since that adds this calculation to the DIII-D mappings

PR #372, opened 2025-08-22
Several bug fixes

AreWeDreaming:
This addresses some miscellaneous bugs that I encountered when fixing the application I will use to test the transition of OMAS to awkward arrays. 
One of those is important because it fixes the `b_field_tor` calculation in OMAS that had a major error in it. Unfortunately, this was not caught by the regression tests of the B-field calculation because it was in the if-branch that is not tested.

### DIFF
--- a/omas/omas_machine.py
+++ b/omas/omas_machine.py
@@ -146,7 +146,8 @@ def machine_to_omas(ods, machine, pulse, location, options={}, branch='', user_m
                         failed_locations[key] = e.TDI
                 except TreeNNF as e:
                     failed_locations[key] = e.TDI
-                    if key != 'equilibrium.time_slice.:.constraints.j_tor.:.measured':
+                    if key not in ['equilibrium.time_slice.:.constraints.j_tor.:.measured',
+                                   'equilibrium.time_slice.:.constraints.j_tor.:']:
                         raise e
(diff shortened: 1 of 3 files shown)


omas/omas_machine.py
```diff
@@ -146,7 +146,8 @@ def machine_to_omas(ods, machine, pulse, location, options={}, branch='', user_m
                         failed_locations[key] = e.TDI
                 except TreeNNF as e:
                     failed_locations[key] = e.TDI
-                    if key != 'equilibrium.time_slice.:.constraints.j_tor.:.measured':
+                    if key not in ['equilibrium.time_slice.:.constraints.j_tor.:.measured',
+                                   'equilibrium.time_slice.:.constraints.j_tor.:']:
```
torrinba, line 150: What's the reason for handling `TreeNNF` differently from `TreeNODATA` and `MdsIpException` here? And why does `j_tor` need special treatment here?

If this is justified, it seems like `j_tor.:.position.phi` would also need to be in the list since that's also one of the `_efit.json` mappings.

AreWeDreaming:
I am merging this because it includes an important fix for the B-field calculation.
@torrinba if you would like to revisit the regression tests please open an issue

PR #377, opened 2025-10-01
Add mappings for equilibrium Jtor and volume profiles

torrinba:
These are fetched from the FLUXFUN tree and interpolated to the EFIT grid.

Also adding a new plotting method that includes the current profile.


torrinba:
Merging so that this can be integrated into the updates for the IRI fetching script

PR #382, opened 2025-11-28
Add mapping for DIII-D RIP diagnostic

torrinba:
This is primarily lifted from IDA_lite.

Setting up the machine mapping to support multiple interferometers and also having RIP couple the polarimeter and interferometer made things a bit more complicated than other mappings.

The changes to nstxu and sample were provided by the format_machine_mapping script.

### DIFF
--- a/omas/machine_mappings/d3d.py
+++ b/omas/machine_mappings/d3d.py
@@ -723,6 +723,7 @@ def interferometer_hardware(ods, pulse):
 
     for i in ods['interferometer.channel']:
         ch = ods['interferometer.channel'][i]
+        ch['wavelength.0.value'] = 10.6e-6 #m
         for field in ch['line_of_sight.first_point'].keys():
             ch['line_of_sight.third_point'][field] = ch['line_of_sight.first_point'][field]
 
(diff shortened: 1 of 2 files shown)


omas/machine_mappings/d3d.py
```diff
@@ -771,6 +772,274 @@ def interferometer_data(ods, pulse):
         )(ods[f'interferometer.channel.{k}.n_e_line.time'])
```
odstrcilt: Just use fill_value="extrapolate", it will do the same. Why is there a negative sign before data[f'{identifier}_validity']?

omas/machine_mappings/d3d.py
```diff
@@ -771,6 +772,274 @@ def interferometer_data(ods, pulse):
...
+    #freq_down = 1000 #Hz
+    #nt = len(time)
+    #n_down = int(round((nt-1)/(time[-1]-time[0]) / freq_down)) // 2 * 2 + 1
+
+    for s in data:
+        offset = data[s][1:ioff].mean()
+        data[s] -= offset
+
+        #use median to keep the fringe jumps sharp
+        #data[s] = np.median(data[s][:nt//n_down*n_down].reshape(-1, n_down), 1)
```
odstrcilt: I'm not sure how well the fringe jump correction work without the downsampling. But we can keep it as it is an improve it if necessary later.

odstrcilt (approved):
Just add the laser wavelength to the IDS and then it should be fine. It it really works as intended will be checked later once IDA is working.

omas/machine_mappings/d3d.py
torrinba: Thanks for noting that. I think that OMAS should serve the data in a general form if possible since there can be other use cases for it. If it is essential to downsample before correcting the fringe jumps we can turn that on (which is why I left the code but commented out) or disable the correction until a better method is implemented.

omas/machine_mappings/d3d.py
torrinba: The IDS for validity uses +1 for certified data, 0 for valid based on automated checks, and <0 for invalid. It looks like the DIII-D `STAT` indicates a problem when it is >1.25 (V) so adding the negative sign is an approximate translation
https://nomos.gat.com/DIII-D/documentation/search.php?search_by=exact&where=signals&searchstring=STATR0&signal_type=PTDATA%20Pointname&tree=N/A

PR #383, opened 2025-12-02
Add mapping for the DIII-D reflectometer diagnostic

torrinba:
This is primarily lifted from IDA_lite.

### DIFF
--- a/omas/machine_mappings/d3d.json
+++ b/omas/machine_mappings/d3d.json
@@ -811,6 +811,7 @@
  },
  "polarimeter.channel.:.wavelength": {
   "PYTHON": "interferometer_polarimeter_hardware(ods, {pulse}, {include_CO2!r}, {include_RIP!r})"
+ },
  "reflectometer_profile.channel.:": {
   "PYTHON": "reflectometer_hardware(ods, {pulse})"
  },


omas/machine_mappings/d3d.py
```diff
@@ -1753,6 +1753,109 @@ def wall(ods, pulse, EFIT_tree="EFIT01", EFIT_run_id=None):
...
+        printe('WARNING: reflectometer data is missing')
+        return
+
+    # assign
+    time = None
+    for k in ods1['reflectometer_profile.channel']:
+        identifier = ods1[f'reflectometer_profile.channel.{k}.identifier'].upper()
+
+        time_ = data[f'{identifier}_time']/1e3 #s
+        freq = data[f'{identifier}_frequency'] #Hz
```
odstrcilt: I do not see where is freq assigned to ods

omas/machine_mappings/d3d.py
```diff
@@ -1753,6 +1753,109 @@ def wall(ods, pulse, EFIT_tree="EFIT01", EFIT_run_id=None):
...
+    Data sources: https://github.com/GA-IDA/ida_lite/blob/c1398c826b7a327d6629b5518c3219b8870436ce/D3D/synt_diags/REFL.py#L41
+
+    :param ods: an OMAS ODS instance
+
+    :param pulse: int
+    """
+    bands=['Q','V','QO', 'VO']
+
+    # reflectometer LOS, not actual location of measured density!
+    R_in, R_out = 1.6, 2.377
```
odstrcilt: 2.377 is from reflectometer documentation, is is the location where the phase is set to be zero.

smithsp:
@odstrcilt In the most recent communication from Suk-Ho about funded diagnostics, reflectometer was listed as not funded.

odstrcilt:
Ok, but it is still essential diagnostic for WEST and it will expected to be in ITER and FPP. I think that we should include it in IDA, even of it will not be availible on DIII-D in the future.

smithsp:
@odstrcilt I agree with the usefulness of historical data.

torrinba:
> it is at 255 degrees according this

Thanks for pointing that out!

PR #393, opened 2026-01-13
Fix for the DIII-D `b_field_tor_vacuum_r` uncertainty

torrinba:
The toroidal field uncertainty was loaded, but not multiplied by the major radius to be consistent with how this variable is defined.


smithsp (approved):
It took me a second to realize how the change related to the description of the PR, but I finally got it.

PR #394, opened 2026-01-14
Fix interferometer mappings, b_field_tor_calc and pyuda imports

AreWeDreaming:
Fixes introduced when working on interferometry and ece


torrinba (approved):
I'm still in favor of renaming this PR for better traceability, but these fixes look good to me and should be merged without delay

torrinba:
@jmcclena can you test that this still works for MAST-U with the `pyuda` changes?

PR #395, opened 2026-01-21
Updated machine mappings and added support_files

cedwardson2:
pegasus.json was updated to fix limiter mappings and add coil current mappings.

Added files to omas/machine_mappings/support_files/pegasus
 - mhdin_p3.dat ~ Updated mhdin file
 - FitWeights.csv ~ Provides fit weights for magnetic probes and current measurements


torrinba:
@cedwardson2 can this be closed now that https://github.com/gafusion/omas/pull/408 was merged?

PR #410, opened 2026-05-20
Fix CER loading

AreWeDreaming:
## Description
The problem with the CER data load is that the `ION.IMPDENS` tree nodes actually point into a different MDSPlus tree (`IMPCON01` and `IMPCON02`). In `getMany` that means that as soon as you hit the first `IMPDENS` node you close the `ION` tree and then never open in again. 

The new code now queries to where the `IMPDENS` nodes are actually pointing and then use regex to formulate an updated query against which ever `IMPCON` flavor was linked in a separate step which resolves the switching.

The resulting fetched data does not agree with imas_composer, but in the spirit of making more PRs that are smaller I would like to track that issue separately.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring

## Pre-Merge Checklist


omas/utilities/omas_mds.py
```diff
@@ -2,6 +2,12 @@
 import os
 from omas.omas_utils import printd
 import numpy as np
+try:
+    import MDSplus
+
+except ImportError:
+    print("Warning no MDSplus! No machine mappings available.")
+    MDSplus = None
```
torrinba: I realize you are a strong proponent of putting the package imports at the top of a script, but this seems like a good example of where it could be better off at the function level. The only new use is in `get_cached_connection` which only gets called from the mappings anyway. There is also another import at the function level already in the `raw` function from the `mdsvalue` class so the pattern isn't new here. I think an error for a missing or undefined package at import in the function would be much more clear than one that says a None type doesn't have a Connection method, which seems like what would happen if `get_cached_connection` somehow did get called without the package being available (which could be pretty far from the `No machine mappings available` warning in the output).

AreWeDreaming: I have pushed a compromise such that the ImportError is reraised should the user attempt to access MDSplus without it being available

omas/machine_mappings/d3d.py
```diff
@@ -1574,28 +1575,53 @@ def charge_exchange_data(ods, pulse, analysis_type='CERQUICK', _measurements=Tru
...
+        if impcon_tree_name is None:
+            impcon_tree_name = tree_name
+        else:
+            assert impcon_tree_name==tree_name, "References to multiple IMCPON trees in one IMPDENS analysis type are not supported."
+        impcon_TDIs[key] = new_path
+        impcon_TDIs[key.replace("_data", "_time")] = f"dim_of({new_path},0)/1000"
     # fetch
     data = mdsvalue('d3d', treename='IONS', pulse=pulse, TDI=TDIs).raw()
-
+    data = data | mdsvalue('d3d', treename=impcon_tree_name, pulse=pulse, TDI=impcon_TDIs).raw()
```
torrinba: This line does not appear to be robust. When loading the machine mappings in OMFIT I get the error:
```
TypeError: unsupported operand type(s) for |: 'dict' and 'dict'
```

torrinba (approved):
I tested the data fetching in OMFIT by hand and it works properly, but I will open a PR to run all the relevant tests for #412 before merging this to be certain

torrinba:
Accidentally merged these in the wrong order... Sorry!

PR #413, opened 2026-05-21
RIP needs a factor 2 because currently it only counts 1 pass

AreWeDreaming:
## Description
When I was working with the RIP diagnostic in IDA it became quickly obvious that the diagnostic was multi-pass but only a single pass was accounted for in `n_e_line`. This PR adds a simple factor of 2 to integrate over both passes.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring

## Testing

Validated against imas_composer, which was used to cross-check the consistency of the CO2 and RIP interferometers in IDA.

## Pre-Merge Checklist

### DIFF
--- a/omas/machine_mappings/d3d.py
+++ b/omas/machine_mappings/d3d.py
@@ -860,7 +860,7 @@ def rip_hardware(ods, pulse):
         z = z + (0,)
     # phi angles are compliant with odd COCOS
     phi0 = 286.6 * (-np.pi / 180.0)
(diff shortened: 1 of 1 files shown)


torrinba (approved):
I've tested by hand that this update still works with fetching in OMFIT.  Since this is a small change that only affects the values returned, not the structure I don't think any further testing is necessary (it seems unlikely that this would have gone unnoticed if the values were actually used in OMFIT)

PR #415, opened 2026-05-21
Very simple addition to have `electrons.density` defined

AreWeDreaming:
## Description
Super simple change to make sure that `electrons.density` is set because looking for `electrons.density_thermal` is unnecessarily complicated.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring

## Testing
Works when imas_composer fetches `electrons.density`

## Pre-Merge Checklist

### DIFF
--- a/omas/machine_mappings/d3d.py
+++ b/omas/machine_mappings/d3d.py
@@ -1916,6 +1916,7 @@ def core_profiles_profile_1d(ods, pulse, PROFILES_tree="OMFIT_PROFS", PROFILES_r
         query = OrderedDict()
         
         # These quantities have an uncertainty associated with them
+        query["electrons.density"] = "N_E"
         query["electrons.density_thermal"] = "N_E"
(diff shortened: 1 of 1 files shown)


torrinba:
It looks like this just makes `density = density_thermal` but IMAS defines it as `density = density_thermal + density_non-thermal`. So are we abusing the definition and if so which one is wrong?

torrinba:
I'm hoping @smithsp can weigh in here. It seems like `density` is a more correct description (if there is one) for the measured quantities, but `density_thermal` could be more correct for modeling applications and is necessary to avoid changes in OMFIT (used in 13 modules and 7 classes) and FUSE. I think it is misleading to provide both with no difference or explanation, but perhaps it is fine to ignore

smithsp:
@torrinba 
Since @AreWeDreaming is an expert in non-thermal electrons, I trust his preference, which is given as this PR.

smithsp (approved):

PR #417, opened 2026-05-22
Bug fix in the DIII-D core_profiles mapping

torrinba:
## Description

This seems like it was an oversight in the core_profiles global quantities mapping function which suggests it has not been sufficiently tested in a while.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring

## Testing

Fetching both OMFIT_profs and ZIPFIT trees in the `imas_composer` tests. Tests that use the default tree options will not trigger this issue, they must be provided in the call for it to be triggered

## Pre-Merge Checklist

- [ ] ~OMFIT has been tested with this OMAS version and you will create a PR on OMFIT that updates the OMAS submodule once this has been merged.~ Not required since unused and only tested with default options.
- [x] Version number has been incremented if this is a major modification or important bug fix


omas/machine_mappings/d3d.json
```diff
@@ -297,7 +297,7 @@
   "PYTHON": "core_profiles_profile_1d(ods, {pulse}, {PROFILES_tree!r}, {PROFILES_run_id!r}, {core_profiles_strict_grid!r})"
  },
  "core_profiles.time": {
-  "PYTHON": "core_profiles_profile_1d(ods, {pulse}, {PROFILES_tree!r}, {PROFILES_run_id!r}, {core_profiles_strict_grid!r})"
+  "PYTHON": "core_profiles_global_quantities_data(ods, {pulse}, {PROFILES_tree!r}, {PROFILES_run_id!r})"
```
AreWeDreaming: @torrinba I don' think this will work. Given that there are two competing functions providing this quantity the d3d.json is just constantly gonna flip flop.

AreWeDreaming (changes requested):
I think we need to dispatch the `core_profiles_global_quantities_data` through a `core_profiles` dispatch function. OMAS needs a 1:1 mapping between quantities and functions. Having the same quantity provided by two different functions that directly arise from the d3d.json causes OMAS to constantly flip-flop.

omas/machine_mappings/d3d.json
torrinba: Good point. I that was an automatic change made by OMAS while I was testing. It can certainly be reverted

AreWeDreaming (approved):
Looks good!

PR #420, opened 2026-06-17
Fix tests

AreWeDreaming:
Pins + fixes to get regression to pass again on `master`.

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -6,30 +6,32 @@ on:
   pull_request:
   workflow_dispatch:
 
+defaults:
+  run:
+    # Login shell so the micromamba environment is activated
+    shell: bash -el {0}
+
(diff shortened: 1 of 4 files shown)


AreWeDreaming:
I cannot reproduce this error locally.

AreWeDreaming:
Using pip I can reproduce this!

AreWeDreaming:
While I can reproduce this now the error is very confusing even with the debugger. The floating point colors come from the matplotlib.rcparams. The strangest part is that if I test this with same matplotlib version but in a pure conda environment then I don't get this bug.
Since our primary use-case for OMAS is to map DIII-D data which needs MDSplus, I'd say we change the CI/CD to install the environment through conda. This has some downstream consequences since omas is distributed on PyPi and then propagated to conda from there (as far as I remember).

PR #427, opened 2026-06-30
Flip sign of j_tor depending on Ip

AreWeDreaming:
Multiply constraint.j_tor.measured with sign(i_p)


torrinba (approved):
Glad this is being addressed! It has probably been overlooked for too long

PR #428, opened 2026-07-14
Add mapping for Zeff

AreWeDreaming:
Mapping for Zeff (see https://github.com/GA-FDP/imas_composer/pull/78)


torrinba (approved):
Nice and straight forward and seems to be working

PR #418, opened 2026-06-16
Extend the time before stale bot acts

torrinba:
## Description

The time frame that stale bot waits before labeling and closing issues and PRs is increased to half a year. This reflects the irregular development that this repo sees.

There appears to be a pretty strong consensus that the stale bot does more harm than good, so I'm beginning to come around to the idea that it may be worth removing completely. I'd welcome any strong opinions on its application to this repo. The changes in this PR could be seen as a compromise or stepping stone between the current bot behavior and complete removal.
https://www.reddit.com/r/programming/comments/sh6a1t/github_stale_bot_considered_harmful/

Also making some tweaks to when the stale bot activates (no one needs to get a notification at 1:30am on a Sunday) and the labels that will deactivate it.

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Code refactoring
- [x] Other (please describe): GitHub Actions

## Testing

N/R only activates after merging

## Pre-Merge Checklist

- [ ] OMFIT has been tested with this OMAS version and you will create a PR on OMFIT that updates the OMAS submodule once this has been merged.
- [ ] Version number has been incremented if this is a major modification or important bug fix
  - To increment the version: Edit the version number in `omas/version` file
  - Follow semantic versioning: MAJOR.MINOR.PATCH (e.g., `0.94.2` → `0.94.3` for bug fixes, `0.94.2` → `0.95.0` for new features)
  - A GitHub release will be automatically created when this PR is merged if the version number has changed

PR #434, opened 2026-08-05
Parse line_id from MDSplus

AreWeDreaming:
## Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Code refactoring

## Testing
Tested against imas_composer https://github.com/GA-FDP/imas_composer/pull/94

## Pre-Merge Checklist


omas/machine_mappings/d3d.py
```diff
@@ -1574,6 +1574,69 @@ def add_n_i_charge_exchange():
...
+
+
+def roman_to_int(roman):
+    total = 0
+    highest = 0
+    for char in reversed(roman):
+        value = ROMAN_DIGITS[char]
+        total += value if value >= highest else -value
+        highest = max(highest, value)
+    return total
```
smithsp, line 1611: Is there not a library that implements this?  

I recommend adding a few tests for this function.

omas/machine_mappings/d3d.py
```diff
@@ -1574,6 +1574,69 @@ def add_n_i_charge_exchange():
     extra_structures["charge_exchange"][f"{sh}.data_error_upper"] = unc_struct
     add_extra_structures(extra_structures)
 
+
+# Element symbol -> (nuclear charge, mass number) for every species CERFIT can fit.
+# D is kept separate from H so that the mass number stays 2.
```
torrinba:
```suggestion
# D is the only uncommon isotope in the list (different mass number than element atomic average for H) so a different letter is used
```

omas/machine_mappings/d3d.py
AreWeDreaming: Do tests that happen inside imas_composer suffice?

omas/machine_mappings/d3d.py
```diff
@@ -1574,6 +1574,69 @@ def add_n_i_charge_exchange():
...
+    if isinstance(lineid, bytes):
+        lineid = lineid.decode()
+    match = LINEID_PATTERN.search(lineid)
+    element = match.group(1) if match else ''
+    roman = match.group(2) if match else ''
+    if element not in CER_ELEMENTS or not roman or not set(roman) <= set(ROMAN_DIGITS):
+        raise ValueError(f'{sub} channel {channel:02d}: cannot parse LINEID {lineid!r}')
+    z_n, a = CER_ELEMENTS[element]
+    z_ion = roman_to_int(roman)
+    symbol = 'H' if element == 'D' else element
```
torrinba, line 1636: The IMAS DD specifically documents that the symbol "D" is allowed. So it doesn't seem like there is a need to change to "H" here.

IMAS also suggests the labels should be "D+", "He+2", "C+6", etc. so it seems like we should try to follow that syntax unless there are problems with it (in which case we should open an issue - this field has been changed to "name" in IMAS 4 so that would be worth keeping in mind there).

omas/machine_mappings/d3d.py
```diff
@@ -1574,6 +1574,69 @@ def add_n_i_charge_exchange():
...
+    'F': (9, 19.0),
+    'Ne': (10, 20.0),
+    'Al': (13, 27.0),
+    'Si': (14, 28.0),
+    'Ar': (18, 40.0),
+    'Ca': (20, 40.0),
+    'Kr': (36, 84.0),
+}
+
+ROMAN_DIGITS = {'I': 1, 'V': 5, 'X': 10, 'L': 50, 'C': 100, 'D': 500, 'M': 1000}
```
torrinba:
```suggestion
ROMAN_DIGITS = {'I': 1, 'V': 5, 'X': 10}
```
Gold plating

omas/machine_mappings/d3d.py
AreWeDreaming: I would defer fixing this for now but open an issue to tackle this later. The reason for this is that we need to also change it in IDA.

torrinba: Why is the label essential in IDA? It seems like `z_n` and `z_ion` are all you need to filter the ions (and `a` if you care about isotopes)

AreWeDreaming: That's not how it is implemented. If the labeling is consistent this is easier: https://github.com/GA-IDA/IDA/blob/2d1b056d9a24fb1262d5b56a69616d3c56d5b251/config/models/charge_exchange.yaml#L19

torrinba: If the label is only used in IDA to distinguish the quantity being measured then it doesn't seem like the format should make any difference. It would only matter if you're parsing the label to extract the ion charge or mass (or have it hard coded anywhere like it is in the config)

AreWeDreaming: Except that right now in IDA this specific label style is also used for the underlying parametrization.

omas/machine_mappings/d3d.py
smithsp: If they test this, that would be sufficient.

omas/machine_mappings/_d3d_cer.py
smithsp, line 1: So all of this got migrated to its own CER omas fetching?

smithsp (approved):

omas/machine_mappings/d3d.py
torrinba: Sure, but the names of the IDA parametrizations don't actually have to match the ion labels and could be selected based on `z_n`, `z_ion`, and `a`

AreWeDreaming: We should try to be consistent

omas/machine_mappings/d3d.py
AreWeDreaming: They do now. https://github.com/GA-FDP/imas_composer/blob/4e730397e2b9fa6fd49d57a9a98b5cb2ed908a24/tests/test_charge_exchange_lineid.py#L67

omas/machine_mappings/_d3d_cer.py
AreWeDreaming: I don't think everything got migrated but I bet that we can find a similar and probably better maintained script somewhere in OMFIT. This script relies on OMFIT in OMAS which is something we are trying to avoid.

AreWeDreaming:
@torrinba can this be merged? It seems like the sister branch in imas_composer was merged which now causes trouble because imas_composer relies on an omas piece provided by this branch for a test the absence of which breaking test discovery

torrinba (approved):
Yeah, this should be fine to merge with the understanding that the ion names are going to change again soon. IMAS will be updating their documentation to recommend the atomic mass, at least for the less common isotopes, with the exception of hydrogen. We don't have to wait for them, but I don't think it needs to be the highest priority yet either

smithsp:
Suspicious that the tests for python 3.7-3.10 pass, but 3.11-3.14 fail.

torrinba:
> Suspicious that the tests for python 3.7-3.10 pass, but 3.11-3.14 fail.

It has been ignored for a while. Probably related to matplotlib changes
https://github.com/gafusion/omas/pull/418#issuecomment-4714050145

PR #439, opened 2026-09-14
Fix bug for Helium beams

AreWeDreaming:
Very small fix that allows support for Helium beams.


torrinba (approved):
Seems like an important fix. Good work

PR #440, opened 2026-09-21
Small fix that adds ece.time

AreWeDreaming:
## Description
Add `ece.time` for d3d (the IDS is homogeneous)

## Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Code refactoring

## Testing
Works for ECRad

## Pre-Merge Checklist

- [x] OMFIT has been tested with this OMAS version and you will create a PR on OMFIT that updates the OMAS submodule once this has been merged. -> Not needed since it is a simple addition
- [x] Version number has been incremented if this is a major modification or important bug fix


omas/machine_mappings/d3d.py
torrinba, line 1277: It looks like this is already being set for every channel. Do you think it's useful enough to have in both places?

omas/machine_mappings/d3d.py
```diff
@@ -1249,6 +1249,7 @@ def electron_cyclotron_emission_data(ods, pulse=133221, fast_ece=False, _measure
             ece_uncertainty[key] = np.sqrt(np.abs(ece_data[key] * 1.e3)) + 70 * np.abs(ece_data[key])
 
     ods['ece.ids_properties.homogeneous_time'] = 0
```
AreWeDreaming: This seems incorrect since there is only one time-source in this entire IDS
```suggestion
    ods['ece.ids_properties.homogeneous_time'] = 1
```

omas/machine_mappings/d3d.py
AreWeDreaming: If `homogeneous_time` is 1 (which I have now added as a suggestion) then we should populate the top level time.

torrinba: It seems like that would justify removing the channel time, except for compatibility with existing use cases. I suppose that is more important

torrinba (approved):
I'm surprised you haven't suggested this change in `imas_composer` as well. Is that forthcoming?

AreWeDreaming:
This was actually for ECRad which still uses Omas

torrinba:
> This was actually for ECRad which still uses Omas

I realize that, but it seems equally relevant to any use of the ECE data

